PR #313, opened 2015-01-06
GIA 3321

jglueck:
GIA-3321 - Ads obscuring articles iOS 8,  similar functionality to Android solution AND-3194. Assets such as images that aren't cached, on a very slow network connection can move the advert placeholder after the coordinates are sent to the native code which actually positions the ad.

The solution runs a timer that checks for any position changes of the placeholder and if it does change triggers an x-gu://ad_moved call which iOS native code then runs updateAdPosition to change the ad's position.


ArticleTemplates/assets/js/modules/ads.js
```diff
@@ -101,43 +101,39 @@ define([
                 });
             },
             getBannerPosCallback : function(callbackNamespace, callbackFunction) {
```
sandropaganotti: callbackNamespace and callbackFunction are defined but not used by the function

ArticleTemplates/assets/js/modules/ads.js
```diff
@@ -101,43 +101,39 @@ define([
...
+            // Android Poll for offSetTop position changes
             posPoller : function(y) {
                 var yPos = y;
-                // console.info('y Loaded position '+yPos);
                 modules.runPoller(1, yPos);
             },
+            // Android Detect if iFrame present
             getAds: function () {
 
                 window.getMpuPosCallback = function (callbackNamespace, callbackFunction) {
```
sandropaganotti: callbackNamespace and callbackFunction are defined but not used by the function

ArticleTemplates/assets/js/modules/ads.js
```diff
@@ -165,6 +159,37 @@ define([
...
+            iosTimer : function(time, yPos) {
+                setTimeout(function() {
+                    modules.runIosPoller(time, yPos);
+                }, 1000);
+            },
+            // iOS Count
+            runIosPoller : function(start, yPos) {
+                var thisNumber = parseInt(start, 10) + 1;
+                var yPolled;
+                if (thisNumber <= 1000) {
```
sandropaganotti: By doing this you're keeping the time running almost forever (1000 seconds) compared to the 20 seconds of the Android version.

ArticleTemplates/assets/js/modules/ads.js
```diff
@@ -177,9 +202,9 @@ define([
                     modules.insertAdPlaceholders(config);
                 }
                 window.getMpuPosJson = modules.getMpuPosJson;
-                window.getMpuPosCommaSeparated = modules.getMpuPosCommaSeparated;
-                window.getBannerPosCallback = modules.getBannerPosCallback;
-                modules.getAds();
+                window.getBannerPosCallback = modules.getBannerPosCallback; // Used by Android
+                modules.getAds(); // Used by Android
+                modules.updateAdsIos(); // Used by iOS
```
sandropaganotti: Can we invoke this function only if we're sure the device is iOs ? Otherwise we'll end up having an extra polling routine running in our page

sandropaganotti:
The next step is, as you suggested to use only one polling function for both Android and iOs.

ArticleTemplates/assets/js/modules/ads.js
jglueck: These were pre-existing parameters I just amended the comment above. However lets remove these and see if everything still works as expected on Android, as it should be ok if not being used.

ArticleTemplates/assets/js/modules/ads.js
jglueck: Same as previous comment.

ArticleTemplates/assets/js/modules/ads.js
jglueck: My own oversite as I placed a big number when initially testing, I've now removed 1000 seconds upper limit and replaced with gradual increase in time for the setTimeOut polling loop, which increments the interval time by 200mls per loop, so the poller runs less frequently the longer the user is on the article (when all assets are most likely to have downloaded even on slow connections).

sandropaganotti:
Ready for code review :)

jglueck:
:+1:  good stuff, we just need an iOS QA to test now with branch GIA-3321 of ios-live and an Android QA test of some articles that contain an interactive and MPU.

sandropaganotti:
Checked, for me is ok. Let's wait until monday to check the iOs part with the iOs team

PR #315, opened 2015-01-09
AND-3342 added Grunt instead of Guard, fixed a few JShint warning

sandropaganotti:
Switched to Grunt instead of Guard. Implemented a mechanism to unify development of android and iOs front-end using rsync.


jglueck:
:+1:

PR #327, opened 2015-02-03
Remove the unused __SUMMARY__ tag from the football liveblog template

rupertbates:
It is never set in the app and is showing up as is on football liveblogs.


sandropaganotti:
:+1: merging into master

PR #355, opened 2015-03-09
AND-3551: Match info not showing in app 

sandropaganotti:


sandropaganotti:
Ready to be code reviewed /cc @guardian/developers-write

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -187,45 +187,45 @@ define([
...
+            });
+
+            bean.on(tabContainer[0], 'click', 'a', function (e) {
+
+                e.preventDefault();                
+                var tab = $(this);
 
+                if( tab.attr("aria-selected") !== 'true' ) {
+                 
                     $($('[aria-selected="true"]').attr('href')).hide();
```
piuccio: You haven't touched this part, but maybe this selector can be cached, `$('[aria-selected="true"]')`

piuccio:
:+1:  looks nice, good to have tests and event delegation

sandropaganotti:
Thanks! ^_^ I've also cached `[aria-selected=true]` as you suggested

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -187,45 +187,46 @@ define([
...
-                    $('[aria-selected="true"]').attr("aria-selected", false);
+                if( tab.attr("aria-selected") !== 'true' ) {
+                 
+                    var activeTab = $('[aria-selected="true"]', tabContainer);
+                    $(activeTab.attr('href')).hide();
+                    activeTab.attr("aria-selected", false);
 
-                    $($(this).attr('href')).show();
-                    $(this).attr("aria-selected", true);
+                    $(tab.attr('href')).show();
```
jamesgorrie: Maybe I am being stupid - but how do you show an attribute of an element?

sandropaganotti: It's because the `href` is in the form `#someid`, so if you do `$(elem.attr('href'))` is like doing `$('#someid')`. It's maybe not so elegant, I agree.

rtyley:
> Ready to be code reviewed /cc @guardian/developers-write

@sandropaganotti - the @guardian/developers-write team is pretty big (67 members), and includes lots of people who don't work on mobile-apps - when you cc them, they all get a notification. Could you use a different team, like @android-developers ?

sandropaganotti:
Sorry @rtyley I'll use guardian/js from now on, thanks for the feedback!

davidrapson:
@rtyley The original context for this was that @sandropaganotti is the only dedicated CS dev on his team so he asked if he could ping other CS devs for review. Agreed the whole dev team is probably too much, but I'd definitely still be interested in seeing the PRs, so if there's a CS-only group to use that would be ideal.

piuccio:
I'd :+1: again, looks good to me

rtyley:
> @sandropaganotti is the only dedicated CS dev on his team so he asked if he could ping other CS devs for review. Agreed the whole dev team is probably too much, but I'd definitely still be interested in seeing the PRs, so if there's a CS-only group to use that would be ideal.

There are 'js' and 'css' teams - I've ensured you're on both now! @sandropaganotti, are you ok to cc those as appropriate?

sandropaganotti:
:+1: yes! That sounds perfect, thanks!

PR #383, opened 2015-06-01
GIA-3675: Fonts are not available within the full screen interactive

sandropaganotti:


GuardianAndroid:

PR #391, opened 2015-06-30
Change message for require('./config') failure

jorgeazevedo:
If there's a syntax error in `config.js`, an empty configuration will
be assumed and "Config module not found" will be printed to the
console.

This patch changes the message to a more generic "failed to load"
explanation and prints out the exception for full detail. This
hopefully makes syntax errors more obvious and easier to find.


GuardianAndroid:

PR #393, opened 2015-06-30
Replace build output with a simple build message

jorgeazevedo:
This patch makes the build stdout to be saved to a .log in each
projects root folder. Errors will still be printed in the console.
It also prints a "Building app" message to show it's doing something.

The main reasoning behind this is to silence the iOS output, as it's
especially verbose and makes it complicated to understand what grunt
is actually doing.


GuardianAndroid:

PR #394, opened 2015-06-30
Update android path to submodule

jorgeazevedo:
The android source tree was re-organized recently. This patch updates
the reference.


GuardianAndroid:

PR #401, opened 2015-07-21
changing absolute paths across static html so that easier to release …

clarehyam:
To ease releasing static html to Microsoft dev team


GuardianAndroid:

PR #421, opened 2015-08-24
Gia 3847

blongden73:
Static templates should all be up to date. Removal of extra comment counter. All the new follow buttons in place. Change of colour to the follow button. We just need to check those CSS changes for the static files haven't broken anything in the actual templates.


GuardianAndroid:

PR #476, opened 2016-04-26
DES-60

GHaberis:
Add support for "personality" quizzes and also refactor quiz code.


ArticleTemplates/assets/js/modules/quiz.js
```diff
@@ -1,251 +1,454 @@
...
+            setupNewsQuizAnswers: function (questionObj) {
+                var i,
+                    answerCode,
+                    answers = questionObj.elem.querySelectorAll('.question__answer');
 
-            // Do we have an MPU and is it below the quiz?
-            var mpu = $('.advert-slot__wrapper').first(),
-                moveMPU = false;
+                for (i = 0; i < answers.length; i++) {
+                    answerCode = String.fromCharCode(65 + i);
```
gtrufitt: Clever! Maybe a comment here to explain that it's looping through alpha a=65,b=66 etc

ArticleTemplates/assets/js/modules/quiz.js
```diff
@@ -1,251 +1,454 @@
...
+                    answer = question.querySelector('[data-correct="true"]');
+                }
+
+                answer.classList.add('correct-answer');
+
+                if (answer.dataset.correctAnswerExplanation) {
+                    answerPara = document.createElement("p");
+                    answerPara.classList.add("answer__explanation");
+                    answerPara.innerHTML = answer.dataset.correctAnswerExplanation.trim();
+                    correctAnswerWrapper = answer.querySelector('.answer__message');
```
gtrufitt: Minor style thing (sorry, I know I would leave code style ;)), you're mixing single and double quotes for strings in this file

ArticleTemplates/assets/scss/modules/_quiz.scss
```diff
@@ -424,5 +440,72 @@
                 }
             }
         }
+
+        // Quiz Results
+        .quiz-results {
+            .quiz-results {
+                &__description {
```
gtrufitt: When I first worked here, I wrote my scss the same, however I was convinced to refactor and write out the classes without splitting in the middle as it made grepping the codebase a lot easier when looking for a specific class. Just a thought.

ArticleTemplates/assets/scss/modules/_quiz.scss
```diff
@@ -424,5 +440,72 @@
...
+
+                    .answer__marker {
+                        border: 1px solid color(shade-1);
+
+                        &::before {
+                            content: '';
+                        }
+
+                        .answer__marker__inner {
+                            opacity: 0;
```
gtrufitt: This is getting pretty nested here, do you need to nest this far down (I imagine you could move `.answer__marker__inner` outside of `answer__marker` at least?

ArticleTemplates/assets/js/modules/util.js
```diff
@@ -41,6 +41,34 @@ define(function () {
 
         isOnline: function () {
             return !document.body.classList.contains('offline') && navigator.onLine;
+        },
```
gtrufitt: At risk of building your own qwery, I wonder whether a `each` util might be useful to clear out some of the for loops and make the code easier to reason about?

gtrufitt:
Looks good dude, there are definitely some learnings we could bring over from dotcom that you're probably already itching to do in time but 👍 from me!

ArticleTemplates/assets/js/modules/util.js
gtrufitt: I also wonder whether it might be worth using something like underscore but enabling you to import only the methods you use (like we do on dotcom), that would probably give you a lot of util methods (though a few possibly superseded by ES2015 if you end up using that...)

ArticleTemplates/assets/js/modules/quiz.js
GHaberis: Will do!

ArticleTemplates/assets/js/modules/quiz.js
GHaberis: Fair point! I'll update to use single quotes for strings for consistency.

ArticleTemplates/assets/scss/modules/_quiz.scss
GHaberis: Yep, I often end up nesting classes at the same level the corresponding node is nested in the dom, which is not really necessary.

PR #491, opened 2016-06-02
DES-70

GHaberis:
Low friction participation component for the app, 5-star review system to be displayed on film reviews.

**bootstraps/common.js**
- add modules/experiments/ab module
- reformat argument names to lower case (consistency)
- remove needless functions/function calls in init

**modules/experiments/ab**
- reads test property in GU.opts.tests and kicks off tests

**modules/experiments/lowFrictionParticipation**
- test logic for low friction star rating component

**shim.js**
- add private methods to private scope, only expose init in  GU.Bootstrap object


GHaberis:
@gtrufitt it's be great if you could take a quick look at this PR.

ArticleTemplates/assets/js/modules/experiments/lowFrictionParticipation.js
```diff
@@ -0,0 +1,263 @@
...
+
+    function clearRating(reviewRating) {
+        var className;
+        var i;
+
+        for (i = reviewRating.classList.length; i > 0; i--) {
+            className = reviewRating.classList[i - 1];
+
+            if (className.indexOf('review-rating--') !== -1) {
+                reviewRating.classList.remove(className);
```
gtrufitt: Could you `break;` this here to avoid looping the remaining classes once you're done?

ArticleTemplates/assets/scss/modules/_lowFrictionParticipation.scss
```diff
@@ -0,0 +1,58 @@
...
+
+    .participation-low-friction__contents {
+    	text-align: center;
+    }
+
+ 	.participation-low-friction__test-message {
+ 		overflow: hidden;
+ 		max-height: 0;
+    	-webkit-transition: max-height 1s linear;
+        transition: max-height 1s linear;
```
gtrufitt: You need some autoprefixer in your build :D

gtrufitt:
Other than the tracking we discussed, looks good to me! 👍

ArticleTemplates/assets/js/modules/experiments/lowFrictionParticipation.js
GHaberis: There's never going to be more than 1 match of 'review-rating--', so yes I'll add a break here.

GHaberis:
@gtrufitt, I've updated the PR following last weeks conversation regarding control/variant in A/B test. So the lowFrictionParticipation test now handles whether the user is in the control bucket or variant bucket and acts accordingly. 

The text also now runs on the comments template as well as film reviews as I need to be able to check whether a user on a comments template whose previously seen the 'control' version of the test goes on to comment.

PR #497, opened 2016-06-20
DES-80 Show placeholder where thumbnails should be when offline and p…

GHaberis:
…ull all image related adjustments together in common.js


GHaberis:
Hi @guardian/dotcom-apps-collab this pull request addresses a bug in the templates, where copy overlays thumbnail image captions when offline, as the image fails to load. 

The updated code replaces the image with a placeholder and sets the dimensions of the placeholder to the would-be dimensions of the image had it displayed.

I've also updated the code a little as there were a number of methods which did various things to images (hideEmptyCaptions, imageSizer, offline and formatThumbnailImages). There was repetition amongst all 4 (continually querying the DOM for the same elements)...I've consolidated common querying into 1 method 'formatImages' which then kicks things off.

PR #561, opened 2016-11-03
Updates to documentation

SiAdcock:
I had a few trip-ups while getting this running locally, so I thought I'd make a couple of updates to the docs:

- make no assumptions about a developer's shell
- set correct paths to native apps' templates directory
- clarify that `ios.sign` and `ios.provisioning` are only needed for testing on a physical iOS device


GHaberis:
@SiAdcock :thumbsup: thanks for doing this!

PR #562, opened 2016-11-03
Remove Hologram

SiAdcock:
Having a dependency on Ruby adds complexity to the project and makes it harder to set up. After speaking to @GHaberis, we thought it would be helpful to get rid of the Ruby dependencies we no longer need.

To start this process, this PR removes Hologram, the Ruby Gem that generates visuals documentation. It looks like much of the documentation hasn't been updated in 2 years, and the `gh-pages` page that hosted the documentation no longer exists. 

If we can confirm that docs are no longer used, I'll start removing the instrumentation from the SCSS. However, do let me know if this documentation is still being used and I'll find a new approach 😄


GHaberis:
@SiAdcock this all looks good to me :thumbsup:

GHaberis:
Hi @SiAdcock! OK I was thinking I'd like to push on and remove the ruby dependencies and build the CSS using node. There's a conflict in the README that needs resolving on this PR but once this is resolved can we go ahead and merge it in?

PR #565, opened 2016-11-23
DES-124

GHaberis:
Add module youtube.js to handle new youtube atoms

Files of note...

**ArticleTemplates/assets/js/modules/youtube.js**
Handles the logic around youtube atoms: Adding script, initialising players, showing/hiding placeholders and reporting tracking events.

**ArticleTemplates/assets/js/bootstraps/article.js**
Initialise youtube module in News Articles

**ArticleTemplates/assets/js/bootstraps/video.js**
Initialise youtube module in Video Articles

**ArticleTemplates/assets/js/bootstraps/liveblog.js**
Initialise youtube module in live blogs, and call youtube.checkForVideos() on new live blog posts and loading more (old) live blog posts.


ArticleTemplates/assets/js/modules/youtube.js
```diff
@@ -0,0 +1,246 @@
...
+            eventType: 'video:content:end'
+        });
+    }
+
+    function showPlaceholder(placeholderParent) {
+        placeholderParent.classList.remove('show-video');
+    }
+
+    function onPlayerPaused(id) {
+        killProgressTracker(false, id);
```
akash1810: why do we do this?

ArticleTemplates/assets/js/modules/youtube.js
```diff
@@ -0,0 +1,246 @@
...
+
+    function getPlaceholder(video) {
+        var placeholder = video.parentNode.getElementsByClassName('youtube-media__placeholder')[0],
+            img = placeholder.getElementsByClassName('youtube-media__placeholder__img')[0];
+
+        if (img.getAttribute('style') !== 'background-image: url()') {
+            return placeholder;
+        } else {
+            placeholder.parentNode.removeChild(placeholder);
+            return false;
```
akash1810: should we have a consistent return type in this function?

ArticleTemplates/assets/js/modules/youtube.js
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    function recordPlayerProgress(id) {
+        var currentTime = players[id].player.getCurrentTime(),
+            percentPlayed = Math.round(((currentTime / players[id].duration) * 100));
+
+        if (percentPlayed > 0 &&
+            percentPlayed % 25 === 0 &&
+            players[id].trackingCalls.indexOf(percentPlayed) === -1) {
```
akash1810: will this fire a `video:content:100` too?

ArticleTemplates/assets/js/modules/youtube.js
GHaberis: @akash I'm going to update this function as I agree getPlaceholder is not totally descriptive of what's going on here.

ArticleTemplates/assets/js/modules/youtube.js
GHaberis: @akash1810 just to prevent the tracker (polling) function from continuing to tick whilst the videos not playing.

ArticleTemplates/assets/js/modules/youtube.js
GHaberis: @akash1810 it will, I can see why this might not be required if we're also firing video:content:end.

gidsg: It appear `100` event is not supported by Ophan, so you shouldn't fire it:
https://github.com/guardian/ophan/blob/1d70ed327d13a0b5dbea4dd0d28f4d0db3fea68f/the-slab/app/extractors/MediaPlaybackExtractor.scala#L38-L41

akash1810: > I'm happy to prevent 100 being fired if this is consistent with other platforms?

Yep, we just fire `end`.

GHaberis: @gidsg @akash1810 OK cool I'll update now so 100 isn't recorded.

ArticleTemplates/assets/js/modules/youtube.js
GHaberis: yes, pretty much although the gains in reality are negligible...It seems like good practice to kill the interval if it's ticking away for no reason.

PR #605, opened 2017-04-12
import util module as dependency

GHaberis:
Refactor modules
- Modules which use the util module no longer refer to window.GU.util. Now they import the util module.

Refactor unit tests
- Remove use of window.GU.util
- Use sandbox for Sinon spy/stub
- Put injector call in beforeEach
- Remove unnecessary done callback if test is sync

Why?
- Removing util from GU namespace means we no longer have to import it into tests, define it in global namespace for tests and clear after each test spec has run. This should hopefully guard against potential race conditions where util is referenced after it's been cleared.


GHaberis:
*drum roll*............it appears to have worked!

PR #607, opened 2017-04-13
football liveblogs need the adsType ‘liveblog’ to be passed to ad module

GHaberis:
Some content types (eg. football) can contain liveblogs, if they do contain an element with the class 'article__body--liveblog' then we need to pass the ads module adsType 'liveblog' to enable ads.

Changes
- We now check if ads enabled in app.js and don't initialise ads module if ads not enabled. I've therefore removed adsEnabled from the object passed to ads.init
- I've created a getAdType function to return correct adType to be passed to ad module.


GHaberis:
Muchas Gracias @SiAdcock !

PR #608, opened 2017-04-25
fix styles in the football article styles

GHaberis:
Fixes some poor formatting introduced by recent changes to the football styles:

- Section 'football' and comment count should have been hidden but were visible
- background of meta area being grey (recent change) looked poor when article beneath it

Before this fix...

img_6236

img_6235

After this fix...

img_6237

img_6238


GHaberis:
Thanks @SiAdcock !

PR #609, opened 2017-05-02
Adding scripts to package.json

NataliaLKB:
Probably my Node.js background, but the first think I do after looking at a README for a project is look at the package.json.

I have filled out the scripts bit to be a bit more helpful (for me).

If you don't want these changes, feel free to close this PR - its just a suggestion.


GHaberis:
Nice, yep anything that helps new comers to the codebase gets a 👍  from me!

PR #611, opened 2017-05-03
Hiding tags if there aren't any

NataliaLKB:
Fixing: https://trello.com/c/NEAp7nks/265-app-template-don-t-show-tags-container-if-no-tags-available

I also took the liberty of changing the classname/css structure a bit just for this change.

I wasn't totally sure if ArticleTemplates/assets/css/style-async.css should be commited. Let me know if I should remove it from this PR.

### After Tags:
image

### Before no tags:
(Notice the space after the article)
image

### After no tags:
(no space)
image


ArticleTemplates/articleTemplateContainer.html
```diff
@@ -43,9 +43,8 @@ <h1 class="headline selectable" id="headline">__TITLE__ __REVIEW_RATING__</h1>
         </div>
 
         <div class="tags" id="tags">
-            <ul class="inline-list" id="tag-list">
-                <li class="inline-list__item screen-readable">Tags:</li>
-            </ul>
+            <span class="screen-readable">Tags:</span>
```
NataliaLKB: Didn't see a reason why this couldn't be span, let me know if there is a reason!

test/fixture/article-tone-analysis.html
```diff
@@ -127,8 +127,11 @@ <h1 class="headline selectable" id="headline">It's only rock'n'roll but they lik
         </div>
 
         <div class="tags" id="tags">
-            <ul class="inline-list" id="tag-list">
-                <li class="inline-list__item screen-readable">Tags:</li>
+            <span class="screen-readable">Tags:</span>
+            <ul class="inline-list tags__inline-list" id="tag-list">
+                <li class="inline-list__item tags__list-item"><a href="#">Guardian tag</a><span class="tags__separator" aria-hidden="true">&#183;</span></li>
```
NataliaLKB: Just added these here so I could visually test this change

ArticleTemplates/assets/scss/modules/_tags.scss
```diff
@@ -46,43 +43,50 @@ Display which keywords an article has been tagged with.
...
+        @include tag-button(color(shade-1), color(global-tags));
+    }
+
+    &.more-button a {
+        background-color: transparent !important;
+        @include tag-more-button(color(shade-1), darken(color(global-tags), 10%));
+    }
+
+    &:last-child {
+        .tags__separator {
```
GHaberis: It looks like the tags__separator style is no longer required as span.tags__separator was removed from articleTemplateTag.html here: https://github.com/guardian/mobile-apps-article-templates/commit/eccf2f335bbf85359c0be2adeb0850832c94b699#diff-64e12bff4e22adfe7c2c5d3a12524316

ArticleTemplates/articleTemplateContainer.html
GHaberis: No I can't think of any reason! This change will however require an update to https://github.com/guardian/mobile-apps-article-templates/blob/master/ArticleTemplates/assets/js/modules/more-tags.js. There's logic in this module to count the tags and only show the first 5 with a 'more' button, which on tap shows the remainder if there are any. This logic takes into account the original "screen-readable" list item jow removed!

ArticleTemplates/assets/scss/modules/_tags.scss
```diff
@@ -46,43 +43,50 @@ Display which keywords an article has been tagged with.
...
-        }
+    @include mq($from: col4) {
+        margin-left: 232px; // 240 - 4 (item padding offset) - 4 (visually offset rounded items);
+        width: 940px;
     }
+}
 
-    .witness + & {
-        margin-top: base-px(2);
+.tags__list-item {
```
GHaberis: Thanks for updating these class names to be syntactically correct BEM names! This will however mean we need to update https://github.com/guardian/mobile-apps-article-templates/blob/master/ArticleTemplates/assets/js/modules/more-tags.js as the old class names are used in the JS. There are some corresponding unit tests for the more-tags module too which will need updating.

NataliaLKB: I have fixed the tests, but you don't really need to change the classnames, as the original classnames are being used too - this just means there is less nesting in (and therefore more performant) css. Let me know if these change are okay

ArticleTemplates/assets/js/modules/more-tags.js
```diff
@@ -14,7 +14,7 @@ define(function () {
             initialised = true;
             tags = document.querySelectorAll('.tags .inline-list .inline-list__item');
 
-            if (tags.length > SHOW_TAGS + 1) {
+            if (tags.length > SHOW_TAGS) {
                 moreButton = document.createElement('li');
                 moreButton.id = 'more-tags-container';
                 moreButton.classList.add('inline-list__item');
```
GHaberis: I think you'll need to add the new class name 'tags__list-item' to moreButton so it picks up the styles.

ArticleTemplates/assets/scss/modules/_tags.scss
GHaberis: Oh yes I missed that the class 'inline-list__item' is still there! Yep, all seems good to me.

ArticleTemplates/assets/js/modules/more-tags.js
```diff
@@ -14,7 +14,7 @@ define(function () {
             initialised = true;
             tags = document.querySelectorAll('.tags .inline-list .inline-list__item');
```
GHaberis: we could actually simplify this selector now to look up any element with the new class name 'tags__list-item'

NataliaLKB: Your right! Good point. Done :)

ArticleTemplates/assets/js/modules/more-tags.js
NataliaLKB: Done 😄

ArticleTemplates/assets/scss/modules/_tags.scss
```diff
@@ -10,9 +10,9 @@ Display which keywords an article has been tagged with.
 <div class="tags">
```
GHaberis: The first 'Tags' list item can be removed from this comment and the class 'tags__list-item' should be added to the other list items in the comment too.

PR #612, opened 2017-05-04
use yarn for dependency management

GHaberis:
Switching from NPM to Yarn for dependency management, in order to get consistent installs across machines and our CI environment.


README.md
```diff
@@ -2,26 +2,27 @@
...
 
 ## Requirements
-* A Mac or Linux PC.
+* A Mac or Linux computer.
 * [brew](http://brew.sh/) as a package manager.
 * [NVM](https://github.com/creationix/nvm).
 * [NodeJS](http://nodejs.org/). Install using nvm: `nvm install v6.9.2`. Remember to add `nvm use v6.9.2` to your preferred shell startup file.
-* [Grunt](http://gruntjs.com/). Install using `npm install -g grunt grunt-cli`.
+* [Yarn] (https://yarnpkg.com). Yarn is a package manager. Install using brew install yarn.
+* [Grunt](http://gruntjs.com/). Install using `yarn global add grunt grunt-cli`.
```
NataliaLKB: If you use the npm scripts you don't need a global grunt. I tend to avoid global dependencies so I prefer this. What do you think?

README.md
```diff
@@ -2,26 +2,27 @@
...
 * It is recommended you restart your shell to ensure changes added the startup file are applied.
 
 ## Usage
 * checkout the project in a separate directory, outside the iOS and the Android app.
-* run `npm install`.
+* run `yarn`.
 * copy config.sample.js to config.js and fill in the details
     * `base.android` is the 'ArticleTemplate' path within the Android app, eg: `/Users/sandropaganotti/Projects/android-news-app/android-news-app/src/main/assets/templates/`
```
NataliaLKB: I found it hard to find these two paths on the Android and IOS projects. Might have just been something silly I did, but could you double check these are correct examples?

README.md
GHaberis: Yeah it makes sense to point user's in the direction of scripts in the package.json rather than running grunt commands on the cli, then if/when we replace grunt user's can still run the same scripts from package.json without caring about the underlying tooling.

So if I remove global grunt/grunt-cli installation from the set-up steps I guess I'll need to add equivalent scripts to package.json for all grunt tasks currently run on the command line where they don't exist? eg. grunt rsync

NataliaLKB: @GHaberis Yep! That sounds completely right. It sounds like that would be an easy job. Let me know if it doesn't work for whatever reason.

README.md
GHaberis: hmm, I cross checked the examples with my config and these seemed ok to me

GHaberis:
@Natlia I've updated readme to remove references to Grunt tasks on command line

README.md
NataliaLKB: Okay cool!

README.md
```diff
@@ -1,35 +1,31 @@
...
-    * `base.ios` is the 'ArticleTemplate' path within the iOs app, eg: `/Users/sandropaganotti/Projects/ios-live/mobile-apps-article-templates/ArticleTemplates/`
+    * `base.ios` is the 'ArticleTemplate' path within the iOS app, eg: `/Users/sandropaganotti/Projects/ios-live/mobile-apps-article-templates/ArticleTemplates/`
     * `base.html` is the path where this repository has been checked out, eg: `/Users/sandropaganotti/Projects/mobile-apps-article-templates/`
 * run `grunt`
```
SiAdcock: `yarn develop`?

SiAdcock:
Love this change, @GHaberis, it will make the barrier to entry so much lower. 

I wonder if we need to list every script in the readme? It feels a bit noisy and overwhelming at the moment, and many of the scripts will not be of interest to developers working with the project. The only scripts we really care about are `develop`, `validate` and `test`, while the others are relevant to CI, or are intermediate scripts that help compose the main scripts.

What do you think?

c/c @NataliaLKB

package.json
```diff
@@ -6,8 +6,12 @@
   "scripts": {
     "test": "grunt karma",
     "validate": "grunt sasslint && grunt jshint",
-    "build": "grunt build",
-    "watch": "grunt watch"
+    "buildJS": "grunt buildJS",
+    "buildCSS": "grunt buildCSS",
+    "build": "grunt buildJS && grunt buildCSS",
```
SiAdcock: We could decouple this from `grunt`:

```json
"build": "npm run buildJS && npm run buildCSS"
```

package.json
```diff
@@ -6,8 +6,12 @@
...
+    "sync": "grunt rsync",
+    "watch": "grunt watch",
```
SiAdcock: Since the `default` grunt task already simply runs `watch`, perhaps we don't need this script?

Gruntfile.js
```diff
@@ -238,7 +238,5 @@ module.exports = function(grunt) {
 
     grunt.registerTask('buildCSS', ['sasslint', 'sass', 'cssmin']);
 
-    grunt.registerTask('build', ['buildJS', 'buildCSS']);
-
-    grunt.registerTask('default', 'watch');
+    grunt.registerTask('default', ['build', 'watch']);
```
SiAdcock: If we're killing the `build` task, will this still work?

GHaberis: good spot, updating now.

PR #614, opened 2017-05-04
Fixup portrait supporting images on tablet

SiAdcock:
## What does this change?

Text ought to wrap around portrait-oriented supporting images.

### Landscape-oriented iPad before

picture 224

### Landscape-oriented iPad after

picture 225

### Portrait-oriented iPad after

picture 226

### No wrap on mobile

picture 227

### DIFF
--- a/ArticleTemplates/assets/scss/layout/_article--immersive.scss
+++ b/ArticleTemplates/assets/scss/layout/_article--immersive.scss
@@ -595,6 +595,10 @@ body.immersive .article--immersive {
                         width: 220px;
                     }
                 }
+
+                + .section__rule {
+                    clear: none;
+                }
             }
 
             &.element--showcase {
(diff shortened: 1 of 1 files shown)


GHaberis:
@SiAdcock this definitely is an improvement. I do think it looks a little wrong having the image start from the very left-hand edge of the viewport, if we remove the negative margin on the figure then it'll align with the paragraphs which I think would look better. Generally I think only full-width elements touch the outer edges of the viewport.

SiAdcock:
Good shout @GHaberis.

I've made the image narrower when in portrait orientation, and also fixed some of the margin alignment issues we discussed IRL

picture 232

NataliaLKB:
@SiAdcock Now you shouldn't have to commit and build files because the CI will build things and then push that change to release and master

NataliaLKB:
@SiAdcock please rebase/merge with master and then run `yarn setup`. You shouldn't be asked to commit build files anymore

PR #616, opened 2017-05-04
Contributions Epic in the liveblog

NataliaLKB:
This PR:
- Adds the placeholder elements for Contributions Epic in liveblog
- Makes sure that the injected element is tracked

Again, wasn't sure if I should include the changed JS in the PR or not (whether it should be commited). Let me know! Its no trouble to commit it and add it to this PR

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/liveblog.js
+++ b/ArticleTemplates/assets/js/bootstraps/liveblog.js
@@ -317,6 +317,7 @@ define([
             window.liveblogTime();
             window.addEventListener('scroll', util.debounce(updateBlocksOnScroll, 100, true));
             liveMore();
+            creativeInjector.trackLiveBlogEpic();
 
             if (GU.opts.isMinute && GU.opts.adsConfig === 'tablet') {
                 minute.init();
--- a/ArticleTemplates/assets/js/bootstraps/liveblog.js
+++ b/ArticleTemplates/assets/js/bootstraps/liveblog.js
@@ -4,14 +4,16 @@ define([
     'modules/youtube',
     'modules/minute',
     'bootstraps/common',
-    'modules/util'
+    'modules/util',
+    'modules/creativeInjector'
 ], function (
     relativeDates,
     twitter,
     youtube,
     minute,
     common,
-    util
+    util,
+    creativeInjector
 ) {
     'use strict';
 
@@ -40,6 +42,17 @@ define([
         } else {
             insertAfterElem.parentNode.insertBefore(block, insertBeforeElem);
         }
+    }
+
+    function checkInjectedComponents() {
+      // check for tweets
+      twitter.checkForTweets();
+
+      // check for youtube video atoms
+      youtube.checkForVideos();
+
+      // if there is a contributions epic, track it
+      creativeInjector.trackLiveBlogEpic();
     }
 
     function liveblogNewBlockDump() {
@@ -76,11 +89,7 @@ define([
 
             window.liveblogTime();
 
-            // check for tweets
-            twitter.checkForTweets();
-
-            // check for youtube video atoms
-            youtube.checkForVideos();
+            checkInjectedComponents();
 
             newBlockHtml = '';
         }
@@ -149,11 +158,7 @@ define([
 
         window.liveblogTime();
 
-        // check for tweets
-        twitter.checkForTweets();
-
-        // check for youtube video atoms
-        youtube.checkForVideos();
+        checkInjectedComponents();
     }
 
     function liveblogTime() {
@@ -312,7 +317,7 @@ define([
             window.liveblogTime();
             window.addEventListener('scroll', util.debounce(updateBlocksOnScroll, 100, true));
             liveMore();
-            
+
             if (GU.opts.isMinute && GU.opts.adsConfig === 'tablet') {
                 minute.init();
             } else {
--- a/ArticleTemplates/assets/js/modules/creativeInjector.js
+++ b/ArticleTemplates/assets/js/modules/creativeInjector.js
@@ -20,16 +20,19 @@ function (
 
     function trackLiveBlogEpic() {
         // if there is already a data-tracked attribute than we don't need to set up tracking again
-        var liveBlogEpicContainer = document.querySelectorAll('.contributions-epic:not([data-tracked])');
+        var liveBlogEpicContainers = document.querySelectorAll('.contributions-epic:not([data-tracked])');
         var liveBlogEpicContainerId;
+        var i;
+        var liveBlogEpicContainer;
 
-        if (liveBlogEpicContainer) {
+        for (i = 0; i < liveBlogEpicContainers.length; i++) {
+            liveBlogEpicContainer = liveBlogEpicContainers[i];
             liveBlogEpicContainerId = liveBlogEpicContainer.getAttribute('id');
 
             liveBlogEpicContainer.setAttribute('data-tracked', 'true');
 
             addEventListenerScroll(liveBlogEpicContainer, liveBlogEpicContainerId);
-        }
+        };
     }
 
     function injectCreative(html, css, id, type) {
--- a/ArticleTemplates/assets/js/modules/creativeInjector.js
+++ b/ArticleTemplates/assets/js/modules/creativeInjector.js
@@ -1,6 +1,6 @@
 define([
     'modules/util'
-], 
+],
 function (
     util
 ) {
@@ -18,8 +18,22 @@ function (
         }
     }
 
+    function trackLiveBlogEpic() {
+        // if there is already a data-tracked attribute than we don't need to set up tracking again
+        var liveBlogEpicContainer = document.querySelectorAll('.contributions-epic:not([data-tracked])');
+        var liveBlogEpicContainerId;
+
+        if (liveBlogEpicContainer) {
+            liveBlogEpicContainerId = liveBlogEpicContainer.getAttribute('id');
+
+            liveBlogEpicContainer.setAttribute('data-tracked', 'true');
+
+            addEventListenerScroll(liveBlogEpicContainer, liveBlogEpicContainerId);
+        }
+    }
+
     function injectCreative(html, css, id, type) {
-        if (util.isOnline() && !document.getElementById(id)) {   
+        if (util.isOnline() && !document.getElementById(id)) {
             injectCSS(css);
             injectHTML(html, id, type);
         }
@@ -27,21 +41,25 @@ function (
 
     function injectCSS(css) {
         var style = document.createElement('style');
-        
+
         style.type = 'text/css';
-        
+
         if (style.styleSheet) {
             style.styleSheet.cssText = css;
         } else {
             style.appendChild(document.createTextNode(css));
         }
-        
+
         document.head.appendChild(style);
+    }
+
+    function addEventListenerScroll(creativeContainer, id) {
+        window.addEventListener('scroll', util.debounce(isCreativeInView.bind(null, creativeContainer, id), 100));
     }
 
     function injectHTML(html, id, type) {
         var creativeContainer = document.createElement('div');
-        
+
         creativeContainer.id = id;
         creativeContainer.classList.add('creative-container');
         creativeContainer.classList.add(type + '-creative-container');
@@ -53,7 +71,7 @@ function (
             injectEpicCreative(creativeContainer);
         }
 
-        window.addEventListener('scroll', util.debounce(isCreativeInView.bind(null, creativeContainer, id), 100));
+        addEventListenerScroll(creativeContainer, id);
     }
 
     function injectInlineCreative(creativeContainer) {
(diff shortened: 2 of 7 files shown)


test/spec/unit/bootstraps/liveblogTest.js
```diff
@@ -140,7 +140,6 @@ define([
             it('sets up scroll event listener', function () {
                 liveblog.init();
 
-                expect(window.addEventListener).to.have.been.calledOnce;
```
NataliaLKB: I don't understand the value of now changing this test to check whether or not its been called twice. That is why I just removed it, rather than replace

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -13,13 +13,28 @@ function (
...
 
+    function trackLiveBlogEpic() {
+      var liveBlogEpicContainer = document.querySelector('contributions-epic');
+      var liveBlogEpicContainerId;
+
+      if (liveBlogEpicContainer) {
+          liveBlogEpicContainerId = liveBlogEpicContainer.getAttribute('id');
+      }
+
+      window.addEventListener('scroll', util.debounce(isCreativeInView.bind(null, liveBlogEpicContainer, liveBlogEpicContainerId), 100));
```
GHaberis: The addEventListener should go inside the if liveBlogEpicContainer block incase no liveBlogEpicContainer/liveBlogEpicContainerId exists. 

Could we also pull out the entire window.addEventListener line and put it inside a function? There's an addEventListener on line 71 we could then replace with a function call as it's exactly the same as this line?

ArticleTemplates/liveblogTemplate.html
```diff
@@ -149,6 +149,8 @@ <h1 class="the-minute__headline selectable">__TITLE__</h1>
             <img src="__TEMPLATES_DIRECTORY__assets/img/loading.gif" alt="Loading…"/>
         </div>
 
+        __CONTRIBUTIONS_EPIC__
```
GHaberis: This __CONTRIBUTIONS_EPIC__ placeholder should be moved out of this template and into liveBlogTemplateBlock.html beneath the div.block element. Keeping it within this template would mean it's always at the bottom of a liveblog, moving it to the block template means it will appear between liveblog blocks.

ArticleTemplates/liveblogTemplateContributionsEpic.html
```diff
@@ -0,0 +1,8 @@
+<div class="contributions-epic" id="__CONTRIBUTIONS_EPIC_ID__">
+    <style class="contributions-epic__styling">
+        // Contributions Epic styling will go here
```
GHaberis: The comments will need to be HTML comments as they'll be within the flow of the HTML

GHaberis: Infact this sits within the Style tags so it'll need to be a CSS comment and the comment on line 6 will need to be an HTML comment.

GHaberis:
This is looking good but i've realised there's one scenario we've missed. The epic appears between liveblog blocks. These blocks can be added to the page after initialisation when liveblog updates are available. This means a creative may not appear on a liveblog initially but could be added when a new block is added, we're not handling this scenario yet.

Theerefore you could expose the function 'trackLiveBlogEpic' in creativeInjector.js and from liveblog.js call trackLiveBlogEpic from the function ;liveblogNewBlockDump'. You can see we do something similar when checking for tweets/youtube atoms in new liveblog blocks.

test/spec/unit/bootstraps/liveblogTest.js
GHaberis: This is fine. Could we add creativeInjector to the mocked dependencies in this test spec? Otherwise when the liveblog tests are run they'll import creativeInjector and go inside this module.

ArticleTemplates/liveblogTemplate.html
NataliaLKB: @GHaberis don't we want it to appear always at the bottom of the liveblog?

ArticleTemplates/assets/js/modules/creativeInjector.js
NataliaLKB: @GHaberis this should be done now

ArticleTemplates/liveblogTemplate.html
NataliaLKB: On further thinking I see why we would want this to be between the blocks instead of at the end so I have moved it. I certainly always thought it was a the end though.

ArticleTemplates/liveblogTemplateContributionsEpic.html
NataliaLKB: Done!

NataliaLKB:
@GHaberis 
> Theerefore you could expose the function 'trackLiveBlogEpic' in creativeInjector.js and from liveblog.js call trackLiveBlogEpic from the function ;liveblogNewBlockDump'. You can see we do something similar when checking for tweets/youtube atoms in new liveblog blocks.

This should be done now.

ArticleTemplates/liveblogTemplate.html
GHaberis: Yeah I don't think we clearly documented what was expected, so that was a fair assumption! I just happen to have seen how it's done on the site (between blocks) and recall conversations about it being between blocks.

test/spec/unit/bootstraps/liveblogTest.js
NataliaLKB: > Could we add creativeInjector to the mocked dependencies in this test spec\

Yes!

> You could then test that functions in the mock are called when expected

Not sure how to test something we don't actually do (replace the placeholder with some html and css) as I don't know how its supposed to look once it happens 😕

ArticleTemplates/assets/js/bootstraps/liveblog.js
```diff
@@ -82,6 +84,9 @@ define([
             // check for youtube video atoms
             youtube.checkForVideos();
 
+            // if there is a contributions epic, track it
+            creativeInjector.trackLiveBlogEpic();
+
```
GHaberis: I should've added earlier that this will also need to go into 'liveblogLoadMore' (called when a user taps the more button to load in old blocks), sorry I didn't think of that before! 

It may be sensible to refactor out twitter.checkForTweets(), youtube.checkForVideos(), creativeInjector.trackLiveBlogEpic() into a function and call that from liveblogLoadMore and liveblogNewBlockDump so we don't have to do these updates in two places.

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -13,13 +13,27 @@ function (
...
+        }
+    }
+
+    function trackLiveBlogEpic() {
+        var liveBlogEpicContainer = document.querySelector('contributions-epic');
+
+        if (liveBlogEpicContainer) {
+            var liveBlogEpicContainerId = liveBlogEpicContainer.getAttribute('id')
+
+            isCreativeInView(liveBlogEpicContainer, liveBlogEpicContainerId)
```
GHaberis: trackLiveBlogEpic is called when a new liveblog blocks are added, however the creative may not be in view at this point, so by calling isCreativeInView(liveBlogEpicContainer, liveBlogEpicContainerId) here directly we'll not track if/when it does eventually scroll into view. We should replace the call to isCreativeInView directly and a listener as below...

window.addEventListener('scroll', util.debounce(isCreativeInView.bind(null, liveBlogEpicContainer, liveBlogEpicContainerId), 100));

NataliaLKB: @GHaberis sorry I must have misunderstood your last comment for this. I will revert

GHaberis: Ah yes sorry, I was saying put the call to _window.addEventListener('scroll', util.debounce(isCreativeInView.bind(null, liveBlogEpicContainer, liveBlogEpicContainerId), 100))_ in a function. Then call this function from line 31.

You could then replace line 56 with a call to this new function as well.

ArticleTemplates/assets/js/bootstraps/liveblog.js
NataliaLKB: I think this is done now :)

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -13,35 +13,60 @@ function (
...
+                trackLiveBlogEpic();
+            } else {
+                window.injectCreative = injectCreative;
+                window.applyNativeFunctionCall('injectInlineCreative');
+            }
+        }
+    }
+
+    function trackLiveBlogEpic() {
+        var liveBlogEpicContainer = document.querySelector('contributions-epic');
```
GHaberis: I must've missed this before but the querySelector needs a '.' prepended to it to select by class name. eg '.contributions-epic'.

You could also update the selector to be '.contributions-epic:not([data-tracked])' which will filter out all those elements already tracked. By doing this we don't need the if block on line 32.

Now that we can potentially have more than 1 epic in a liveblog we should also use querySelectorAll as querySelector only returns the first matching element.

ArticleTemplates/assets/js/bootstraps/liveblog.js
```diff
@@ -312,7 +317,8 @@ define([
             window.liveblogTime();
             window.addEventListener('scroll', util.debounce(updateBlocksOnScroll, 100, true));
             liveMore();
-            
+            creativeInjector.init();
```
GHaberis: Now we have the 'trackLiveBlogEpic' exposed from creativeInjector we could actually just call this directly rather than call the 'init'.

By doing this we can remove the logic in creativeInjector.init that checks whether it's a liveblog.

ArticleTemplates/liveblogTemplateBlock.html
```diff
@@ -8,4 +8,7 @@
         </div>
         __BLOCK_LAST__
     </div>
+
+    __CONTRIBUTIONS_EPIC__
```
GHaberis: This placeholder should go beneath div.block, as the creative will be a sibling of div.block.

test/spec/unit/bootstraps/liveblogTest.js
GHaberis: Thanks for adding it. I was thinking you could have an expecatation that creativeInjectorMock. trackLiveBlogEpic is called when a new liveblog block is added and more liveblog blocks are loaded. eg. we check whether twitterMock.checkForTweets has been called.

expect(twitterMock.checkForTweets).to.have.been.calledOnce;

We don't need to check anything beyond this as that's the responsibility of the creativeInjector module and would sit in the unit tests for that module...

NataliaLKB:
@GHaberis Thanks for your helpful explanations and review! Let me know if there is anything else I should change :)

ArticleTemplates/assets/js/bootstraps/liveblog.js
```diff
@@ -312,7 +317,7 @@ define([
             window.liveblogTime();
             window.addEventListener('scroll', util.debounce(updateBlocksOnScroll, 100, true));
             liveMore();
-            
+
```
GHaberis: We'll still need to call creativeInjector.trackLiveBlogEpic() from the ready function incase any are in the page from initialisation, calling it after liveMore() should be fine.

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -18,30 +18,48 @@ function (
         }
     }
 
+    function trackLiveBlogEpic() {
+        // if there is already a data-tracked attribute than we don't need to set up tracking again
+        var liveBlogEpicContainer = document.querySelectorAll('.contributions-epic:not([data-tracked])');
```
GHaberis: document.querySelectorAll will return a node list, so we should rename liveBlogEpicContainer to liveBlogEpicContainers as it may have multiple elements in it.

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -18,30 +18,48 @@ function (
...
+        var liveBlogEpicContainerId;
+
+        if (liveBlogEpicContainer) {
```
GHaberis: we should replace this if block with a for loop as liveBlogEpicContainers will be a list potentially containing multiple epic HTML elements, we'll need to set up tracking on each element in this node list. eg.

```
var i;
var liveBlogEpicContainer;

for (i = 0; i < liveBlogEpicContainers.length; i++) {
    liveBlogEpicContainer = liveBlogEpicContainers[i];
    liveBlogEpicContainerId = liveBlogEpicContainer.getAttribute('id');
    liveBlogEpicContainer.setAttribute('data-tracked', 'true');
    addEventListenerScroll(liveBlogEpicContainer, liveBlogEpicContainerId);
};

```

NataliaLKB:
Thanks @GHaberis ! Really appreciate all your guidance.

PR #620, opened 2017-05-10
Adding a docs repo

NataliaLKB:
Starting off a docs directory.

AND ... this PR should build 🎉 

# Circle CI branch builds:
I have configured the Circle to build all PRs (not all branches, only branches which are open PRs). This means the default branch (master) will build all commits associated with it as well.

To make sure we can still release, I will change master build to also "deploy" (run the ./circle.sh file, but change it so that it pushes the release to the `release` branch.

How does that sounds @GHaberis ?


circle.yml
```diff
@@ -8,17 +8,12 @@ dependencies:
...
-      - release
-
 test:
   override:
     - yarn build
 
 deployment:
   release:
-    branch: release
+    branch: master
```
NataliaLKB: I am pretty sure this will work already given `circle.sh` already commits and pushes to both `master` and `release`

Lets test it @GHaberis 😄

NataliaLKB:
@GHaberis I am merging this to test it.

PR #621, opened 2017-05-10
Ads and epics shouldn't be inserted after the same block

NataliaLKB:
This PR: 
* Changes the name of the class to not conflict with a same already in the markup submitted by the contributions team
* If there is a contributions epic after a liveblog block which an ad should go, the ad should instead go after the next liveblog block.

### DIFF
--- a/ArticleTemplates/assets/js/modules/ads.js
+++ b/ArticleTemplates/assets/js/modules/ads.js
@@ -39,17 +39,19 @@ function (
     }
 
     function updateLiveblogAdPlaceholders(reset) {
-        var i,
-            advertSlots,
-            mpu,
-            block,
-            blocks = document.querySelectorAll('.article__body > .block');
+        var i;
+        var advertSlots;
+        var mpu;
+        var block;
+        // The selector here is taking all .block elements within article body
+        // which are not siblings of contributions-epic__container
+        var blocks = document.querySelectorAll('.article__body > .block:first-child, .article__body > div:not(.contributions-epic__container) + .block');
 
         if (reset) {
             advertSlots = document.getElementsByClassName('advert-slot--mpu');
 
-            for (i = advertSlots.length; i > 0; i--) {
-                advertSlots[i-1].parentNode.removeChild(advertSlots[i-1]);
+            while(advertSlots.length > 0){
+                advertSlots[0].parentNode.removeChild(advertSlots[0]);
             }
 
             numberOfMpus = 0;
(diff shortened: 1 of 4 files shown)


ArticleTemplates/assets/js/modules/ads.js
```diff
@@ -39,31 +39,34 @@ function (
...
+        var advertSlots;
+        var mpu;
+        var block;
 
         if (reset) {
             advertSlots = document.getElementsByClassName('advert-slot--mpu');
 
-            for (i = advertSlots.length; i > 0; i--) {
-                advertSlots[i-1].parentNode.removeChild(advertSlots[i-1]);
+            while(advertSlots.length > 0){
```
NataliaLKB: I find this much easier to reason about. Hope thats okay!

ArticleTemplates/assets/js/modules/ads.js
```diff
@@ -58,12 +60,12 @@ function (
...
             block = blocks[i];
 
-            if (i === 1 || i === 6) {
+            if (i === 2 || i === 7) {
                 numberOfMpus++;
                 mpu = createMpu(numberOfMpus);
 
                 if (block.nextSibling) {
-                    block.parentNode.insertBefore(mpu, block.nextSibling);
+                    block.parentNode.insertBefore(mpu, block);
```
NataliaLKB: I also found this easier to reason about

test/spec/unit/modules/adsTest.js
```diff
@@ -263,7 +263,7 @@ define([
                 };
             });
 
-            it('inserts liveblog ads after 1st and 7th blocks', function (done) {
+            it('inserts liveblog ads after 2nd and 7th blocks', function (done) {
```
NataliaLKB: I am pretty sure it was supposed to be after the 2nd

ArticleTemplates/assets/js/modules/ads.js
GHaberis: This will result in advert's going after the 3rd and 8th blocks, they should still go behind the 2nd and 7th. 

The only time they shouldn't is if an epic is between the 2nd/3rd and 7th/8th...then they should go between the next block immediately proceeded by a block.

test/spec/unit/modules/adsTest.js
GHaberis: yep that's a typo, thanks!

ArticleTemplates/assets/js/modules/ads.js
NataliaLKB: They do still go after the 2nd and 7th block. See this test: https://github.com/guardian/mobile-apps-article-templates/pull/621/files/a45f307dd8db9c79178f1207d3a51f03778fbc12#diff-dcf2f2e661a79bd5a44f16f89b56608aR269

Unless I have misunderstood the test

GHaberis: Ok great I see how it works with the new querySelectorAll If I'm correct this is how it'd add an Advert (A) if there's a epic (C) behind a block (B): 

BBBBBBBB ----> BB**A**BBBBBB
BCBBBBBBB ----> BCBB**A**BBBBB**A**B
BBCBBBBBB ----> BBCB**A**BBBBB**A**B

This looks good to me!

PR #636, opened 2017-05-25
Cross-origin messaging for atoms

regiskuckaertz:
Largely inspired by the work done in `frontend` and `commercial-templates`, this introduces the ability for an embedded atom to communicate with the parent article.

I already got the resizing working and need now to focus on passing webfonts down; just wanted to raise the PR now so you can already start looking for code quality and correctness.


ArticleTemplates/assets/js/modules/messenger.js
```diff
@@ -0,0 +1,205 @@
...
+              continue;
+            }
+
+            var matches = iframes[i].src.match(hostRX);
+            if (!matches) {
+              continue;
+            }
+
+            allowedHosts.push(matches[1] + '://' + matches[2]);
+            iframes[i].contentWindow.postMessage('go', '*');
```
regiskuckaertz: Embedded iframes will be waiting for this message before sending anything. Now it is not 100% foolproof. Depending on the vagaries of network conditions, the embedded document may not be listening yet... I need to improve that part.

ArticleTemplates/assets/js/modules/messenger.js
```diff
@@ -0,0 +1,198 @@
+/* global Promise */
```
GHaberis: @regiskuckaertz unfortunately we have to support Android 4.1 onwards and Promises aren't supported by Android until 4.4.4 http://caniuse.com/#feat=promises :confounded:

GHaberis: I've just checked and only 11% of our Android userbase are running < 4.4.4. Perhaps we can make the call that as this only likely get smaller we can take this hit.

regiskuckaertz: 😭  one question I wanted to ask you is the policy around support strategy. This is your call, I can work around that by using callbacks instead if you feel it is safer.

ArticleTemplates/assets/js/modules/messenger.js
```diff
@@ -0,0 +1,198 @@
...
+    };
+
+    function start(modules) {
+        modules.forEach(function (module) {
+            module.init(register);
+        });
+
+        var iframes = document.getElementsByTagName('iframe');
+        for(var i = 0, ii = iframes.length; i < ii; i++) {
+            if (!iframes[i].classList.contains('atom-embed')) {
```
regiskuckaertz: Yep

ArticleTemplates/assets/js/modules/messenger.js
GHaberis: The strategy to data has been to hold back on using 'newer' features of JS. However I'd be keen to move away from that and look at options that allow us to use these features, for example we could look at polyfilling, it may be that we do this on-demand for now and polyfill Promises. I'd be happy with this if we want to ensure full-support.

ArticleTemplates/assets/js/modules/messenger/resize.js
```diff
@@ -0,0 +1,39 @@
...
+
+        if (specs.width) {
+            styles.width = normalise(specs.width);
+        }
+
+        if (specs.height) {
+            styles.height = normalise(specs.height);
+        }
+
+        Object.assign(iframe.style, styles);
```
GHaberis: Unfortunately Object.assign isn't supported in Android

ArticleTemplates/assets/scss/style-sync.scss
```diff
@@ -22,6 +22,7 @@
 @import 'modules/loading';
 @import 'modules/screen-readable';
 @import 'modules/selectable';
+@import 'modules/storyquestions';
```
GHaberis: You can just include this in style-async.scss. style-sync.scss is for render-blocking styles generally affecting elements above the fold. style-async.css is added subsequently for below the fold elements, like the story questions.

GHaberis:
@regiskuckaertz the logic within messenger/resize all looks good! 👍  let's chat about a course of action regarding promises and unsupported features.

regiskuckaertz:
Just added another module to exchanges webfonts, in the hope that we can remove it later by using the capture facility of mobile apps. It is working, although my web server did not allow CORS requests:

screen shot 2017-05-30 at 11 30 34

ArticleTemplates/assets/js/modules/messenger/resize.js
regiskuckaertz: No worries, I'll unpack this.

ArticleTemplates/assets/js/modules/messenger.js
regiskuckaertz: I guess there are two options:
- polyfill.io: only clients with no support for Promise will get something back. The apparatus that Alex and Simon had to put in place to make this work is quite intricate, I don't know if it is a good idea.
- a bundled polyfill: works everywhere although affects all clients regardless of their level of support.

GHaberis: We need to support offline reading so we'd have to bundle the polyfill with the assets served from the device. This is something I'm going to look at this afternoon as this PR is dependant on the bundled polyfill being available.

GHaberis: @regiskuckaertz this PR will mean Promise is available https://github.com/guardian/mobile-apps-article-templates/pull/637

regiskuckaertz: 🎉

regiskuckaertz:
@GHaberis if you can have another look at the `webfonts` module, that would be lovely

GHaberis:
@regiskuckaertz I've just been looking at webfonts. I ran a test on device attempting to access document.styleSheets[0].cssRules returns null. I believe it's linked to this issue https://bugs.chromium.org/p/chromium/issues/detail?id=143626#c11 as the stylesheets are served from the file:// protocol.

regiskuckaertz:
argh, I've been using `python -m SimpleHTTPServer` to serve the project folder on `http://localhost:8000`. Would you mind trying that way?

regiskuckaertz:
just checked and the `File` interface cannot be used with `file://` either 😭  there was a filesystem API proposal but it has been taken out of the standards track. I don't have much choice but to bundle the webfonts inside the iframe.

GHaberis:
:disappointed: Ok we can put the font sharing on hold for now, but this is something I'm keen to investigate.

GHaberis:
I guess we can remove the maybe.js for now too.

test/fixture/article-tone-analysis.html
```diff
@@ -124,6 +124,7 @@ <h1 class="headline selectable" id="headline">It's only rock'n'roll but they lik
                 <p>“But popularity and sellability is based on creating desire, emotion and passion – that’s what Saint Laurent is about. Given that so many critics themselves just walk around in simple coat and trousers you have to wonder, what is wrong with simplicity?”</p>
             </div>
             <span id="tell-me-when-placeholder"></span>
+            <iframe class="atom-embed storyquestions" name="hello" id="hello" src="http://localhost:9000/embed/atom/storyquestions/b0eba1d8-479f-46a0-b19d-4d94a5f3d728"></iframe>
```
GHaberis: Can we remove this test story question atom can be removed from the fixture?

regiskuckaertz: Yep, that was a mistake.

regiskuckaertz:
Ok I have bundled the webfonts:

screen shot 2017-06-02 at 16 26 13

Only the serif ones though, not the sans, because that's too much to just display a confirmation message.

PR #639, opened 2017-06-05
CI merge conflict fix

GHaberis:
On the CI env we merge master into release, we should always take the changes from master over release if/when we have conflicts.

### DIFF
--- a/circle.sh
+++ b/circle.sh
@@ -13,7 +13,7 @@ if [[ `git status --porcelain` ]]; then
     git commit -m "Generate files for release"
     git checkout release
     git reset --hard origin/release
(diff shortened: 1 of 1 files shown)


circle.sh
```diff
@@ -13,7 +13,7 @@ if [[ `git status --porcelain` ]]; then
     git commit -m "Generate files for release"
     git checkout release
     git reset --hard origin/release
-    git merge master -m "Merge master into release"
+    git merge -X theirs master -m "Merge master into release"
```
NataliaLKB: Wow! I hadn't seen this before. Cool 💯

PR #645, opened 2017-06-09
Fix meta font size

SiAdcock:
There is an [interesting feature of iOS Safari](https://stackoverflow.com/questions/3226001/some-font-sizes-rendered-larger-on-safari-iphone) that resizes text if the browser feels the font size will be too small.

This is causing the timestamp and author follow link text to appear larger than expected in iOS 10.3

I have fixed this by applying the property `-webkit-text-size-adjust: none;` to the `meta` mixin.

## Screenshots

Before:

picture 244

After:

picture 243

### DIFF
--- a/ArticleTemplates/assets/scss/helpers/_mixins.scss
+++ b/ArticleTemplates/assets/scss/helpers/_mixins.scss
@@ -8,10 +8,10 @@
     $headline-lead-3: 4.8rem;
     $headline-size-4: 3.6rem;
     $headline-lead-4: 3.6rem;
-    
+
     font-size: $headline-size-1 * $multiple;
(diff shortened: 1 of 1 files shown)


ArticleTemplates/assets/scss/helpers/_mixins.scss
```diff
@@ -123,6 +123,8 @@ $meta-lead-2: 2.2rem;
         font-size: $meta-size-2 * $multiple;
         line-height: $meta-lead-2 * $multiple;
     }
+
+    -webkit-text-size-adjust: none;
```
SiAdcock: ^ This is the substantial change

PR #648, opened 2017-06-28
Stop letters from touching in immersive template

duarte:
This increases line-height in immersive headlines, and tightens spacing around the headline so the header doesn't take as much space.

## Before & After

animated


GHaberis:
@duarte do you want to merge this?

duarte:
Yeah — are there any additional steps to do after merging or shall I just hit the button?

PR #651, opened 2017-07-07
Send the path of the article in the handshake response

regiskuckaertz:
That is necessary for reporting data back to Ophan, as the iframe origin is on a different domain.


tomrf1:
👍

regiskuckaertz:
Closing this as we found a more efficient solution

PR #652, opened 2017-07-10
Add source maps to CSS during development

GHaberis:
If you run 'yarn develop' you now get source maps with your CSS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "buildCSS": "grunt buildCSS",
     "build": "yarn buildJS && yarn buildCSS",
     "sync": "yarn build && grunt rsync",
-    "develop": "grunt --sourceMap=true"
+    "develop": "grunt"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "buildCSS": "grunt buildCSS",
     "build": "yarn buildJS && yarn buildCSS",
     "sync": "yarn build && grunt rsync",
-    "develop": "grunt"
+    "develop": "grunt --sourceMap=true"
```
NataliaLKB: Can sourceMap always be true?

GHaberis: Yep we could remove the flag and always have source maps. The reason I put it in however was that I didn't want our CI build generating source maps that'd then be bundled onto the device. However I've taken a look at the iOS/Android build scripts and both exclude maps from their bundle so this isn't actually a problem. 

I'll update so source maps are always generated.

NataliaLKB: Great :)

PR #653, opened 2017-07-11
Support display hint 'photoEssay'

GHaberis:
Add style overrides if the class display-hint--photoEssay is on the body (this class will be added by the native layer if an article has the property displayHint with the value photoEssay in it's MAPI response).

The main changes are:
- Half width elements (images and blockquotes) are styled as expected.
- Figcaptions are hidden and if an unordered list follows an image this is used as the caption instead.

Other changes:
- Previously we'd hide empty figcaptions with JS, this was visible to the user, I've updated to do this with CSS and it's no longer visible. 

**Old look photo essay in app...**

photo-essay-bugs

**New look photo essay in app...**

photo-essay-iphone


alfavata:
I've tested this branch with my client-side changes (see https://github.com/guardian/ios-live/pull/3347) and there's a couple of issues...

img_6865
img_6864

GHaberis:
@alfavata thanks for feedback. I've updated the styles to fix these issues, would you mind testing against your branch to confirm it's ok?

duarte:
@GHaberis do you have a link to this essay on the web? I'm wondering if in addition to all the changes you've already implemented, the captions are also meant to look different

GHaberis:
@duarte here it is: https://www.theguardian.com/artanddesign/2017/jun/22/a-hot-summer-night-in-london-photo-essay

You'll see there are a number of differences, notably web supports a full screen hero image and the figcaptions look completely different to standard fig captions 

When I picked this up it was to address the reported 'bugs':

- List item's not being used as captions, so you'd see the original caption and an un-styled list beneath the image with the caption repeated.
- Half-width images were showing full width
- Showcase images were unstyled

I've fixed all the above points but not proceeded with a full implementation of the 'photoEssay' display hint, as I wanted these improvements to be released asap.

GHaberis:
This branch can't be merged until the Android work is ready, the iOS changes have been made and are dependant on this branch and therefore Android.

alfavata:
@GHaberis Android work done and merged

SiAdcock:
Do we need to update [`test/spec/unit/bootstraps/articleTest.js`](https://github.com/guardian/mobile-apps-article-templates/blob/99b1d3adb9079d56fb19ca304c709f65374f1ddb/test/spec/unit/bootstraps/articleTest.js#L47-L51), which is still setting a value for `isImmersive`?

PR #657, opened 2017-07-25
update docs to reflect changes to iOS releases

GHaberis:
The mobile-apps-article-templates NPM package is now used by the ios-live repo. These changes update the docs to reflect this change.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -52,7 +52,7 @@ module.exports = function(grunt) {
     });
 
     grunt.initConfig({
-        // sync templates to local ios/android projects
+        // sync templates to local android projects
         rsync: {
             options: {
                 recursive: true,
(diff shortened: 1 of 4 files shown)


jorgeazevedo:
Cool! Unfortunately the documentation will have to be a bit complicated until we move Android to npm. I'm hoping we can get that in place this week.

Note, this paragraph is rendering as a whole block rather than individual sections

mobile-apps-article-templates_how-to-deploy_md_at_31bedbf83ca280d070e3c181cfaec30640580437_ _guardian_mobile-apps-article-templates_

GHaberis:
@jorgeazevedo yep unfortunately for now there'll be 2 different sets of instructions for iOS/Android but hopefully we'll unify them soon! Thanks for pointing out formatting issue, I've fixed that.

PR #668, opened 2017-09-11
Fix issue with byline being hidden in paid for articles

GHaberis:
The Meta container was being removed on paid for content in the app even when there was a byline.

This was happening because the logic didn't take into account that  `bylineElems[0].children.length` could be 0 even if there was text content within `bylineElems[0]`. The original logic was written to accommodate bylines wrapped in anchor tags that linked to the author profile, however not all are wrapped in anchor tags.

I've updated the check to `bylineElem.innerHTML === ""` which fixes the issue.

I've also simplified this logic, as the only tone ever present on paid for content is `tone-news` so the a big chunk of the previous logic was totally redundant.

Before...

picture 70

After...

picture 69

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/common.js
+++ b/ArticleTemplates/assets/js/bootstraps/common.js
@@ -583,54 +583,29 @@ define([
     }
 
     function advertorialUpdates() {
-        var tones, tone, type, 
-            parentNodeClass, bylineElems, 
-            elemsToDelete, j;
-
-        tones = {
-            'tone--media': {
-                'video': 'meta__misc',
-                'gallery': 'meta__misc',
-                'audio': 'byline--mobile'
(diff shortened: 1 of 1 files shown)


ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -583,54 +583,29 @@ define([
...
-            },
-            'tone--news': 'meta',
-            'tone--feature1': 'meta',
-            'tone--feature2': 'meta',
-            'tone--feature3': 'meta',
-            'tone--podcast': 'byline--media'
-        };
-
         if (GU.opts.isAdvertising) {
-            for (tone in tones) {
```
NataliaLKB: Oh my. So happy this nesting is gone.

PR #673, opened 2017-09-20
Rendering atoms using the library

regiskuckaertz:
This is to insert the CSS and JS necessary for atoms to run properly. CSS is straightforward. For JS, it is trickier. We need a reference to the generated JS, as each atom will expose a factory module. To avoid polluting the global scope, atoms will be assigned to the `this` object (each JS bundle will be of the form `this.<atom type> = <expression returning a module>`), which we can define by wrapping the generated code inside a function and call that function using `Function.prototype.call`, passing an object to serve as a repository.

TODO:
- [x] kick-off atoms in `GU.bootstrap.init`
- [x] safety-check the shape of the atom factory and its result
- [x] provide interface to native implementation of services (ophan, identity, etc.)

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/common.js
+++ b/ArticleTemplates/assets/js/bootstraps/common.js
@@ -33,7 +33,6 @@ define([
         comments.init(); // load comments
         cards.init(); // load cards
         loadEmbeds();
-        // smoothScroll.init(); // scroll to anchor
         loadInteractives(); 
         setupOfflineSwitch();
         setupAlertSwitch();
@@ -47,6 +46,10 @@ define([
         setupTracking(); // track common events
         ab.init(); // init ab tests
 
+        if (smoothScroll !== undefined && typeof smoothScroll.init === 'function') {
+            smoothScroll.init(); // scroll to anchor
+        }
+        
         if (!document.body.classList.contains('no-ready')) {
             util.signalDevice('ready');
         }
--- a/ArticleTemplates/assets/js/bootstraps/common.js
+++ b/ArticleTemplates/assets/js/bootstraps/common.js
@@ -33,7 +33,7 @@ define([
         comments.init(); // load comments
         cards.init(); // load cards
         loadEmbeds();
-        smoothScroll.init(); // scroll to anchor
+        // smoothScroll.init(); // scroll to anchor
         loadInteractives(); 
         setupOfflineSwitch();
         setupAlertSwitch();
--- a/ArticleTemplates/articleTemplate.html
+++ b/ArticleTemplates/articleTemplate.html
@@ -21,6 +21,11 @@
     __ARTICLE_CONTAINER__
 
     <script type="text/javascript">
+        var atoms = {};
+        function readAtoms() {
+            __ATOMS_JS__
+        }
+        readAtoms.call(atoms);
         GU.bootstrap.init({
             sectionTone: "__SECTION_TONE__",
             isImmersive: "__IS_IMMERSIVE__" ? true : false,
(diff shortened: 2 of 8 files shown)


GHaberis:
@regiskuckaertz this looks great!

Would it be safe to add checks to confirm the injected atomFactory and the object it returns contain the methods we attempt to execute in atoms.js [here](https://github.com/guardian/mobile-apps-article-templates/pull/673/files#diff-084e9476e1d26dd33b3b76df956621a0R21) and also [here](https://github.com/guardian/mobile-apps-article-templates/pull/673/files#diff-084e9476e1d26dd33b3b76df956621a0R31)?

regiskuckaertz:
@GHaberis thanks 😄  Yes I definitely need to add that! It's working:

screen shot 2017-09-20 at 14 30 07

🎉

GHaberis:
@regiskuckaertz quick question regarding the `services` import. Correct me if I'm wrong, but I'm interpreting this as the `services` module exports an interface like `{ ophan, viewport }`. 

`services.ophan` will have a method `record`. This method will accept an object as an argument. The app's implementation of the record method will pass that object to the native layer to be tracked in ophan.

`services.viewport` will have a method `observe`. This method will accepts 3 arguments:
 
- The HTML element to observe
- The threshold (a number between 0 and 1 for describing how visible the elements should be)
- A callback to execute once that threshold has been reached

Like so... 

`services.viewport.observe(elem, 0.5, function() { services.ophan.record({}) });`

regiskuckaertz:
@GHaberis Yes, exactly.

regiskuckaertz:
The only reason I had `observe` return an object is that it allows to track multiple elements with the same observer; this is how `IntersectionObserver` works. Though I'm not sure we'll need it, perhaps in some future iteration where there is some shared state between atoms 😄

regiskuckaertz:
@GHaberis I've noticed buttons et al inherited styles from the user agent:

screen shot 2017-09-27 at 16 58 31

This might explain why there are discrepancies between snippets in frontend and in apps. Would it be cool to reset them in a separate PR?

GHaberis:
@regiskuckaertz sounds good to me 👍

GHaberis:
@regiskuckaertz yes that seems like an sensible suggestion to me

regiskuckaertz:
Just added an implementation for the viewport service. I had to add the ratio to the signature of `unobserve`, it made things more tractable.

SLambrakis:
Tested on iOS and all is working well.

regiskuckaertz:
giphy

GHaberis:
@regiskuckaertz, I'll be giving this a 👀  again today. I'm guessing you're happy to merge as soon as it's got a 👍 ?

regiskuckaertz:
@GHaberis I think so, yes. I've tested both implementations of the `viewport` service and they do work fine. The code is sufficiently battle-tested afaik, now I'm keen to see how it will behave in the wild.

NB: I've commented out `smoothScroll` as it seems to break in iOS Safari 11; you might want to do something about this.

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -33,7 +33,7 @@ define([
         comments.init(); // load comments
         cards.init(); // load cards
         loadEmbeds();
-        smoothScroll.init(); // scroll to anchor
+        // smoothScroll.init(); // scroll to anchor
```
GHaberis: So this is causing issues on iOS 11? This library is primarily used for scrolling the related cards at the bottom of an article. What happens when it's enabled? Does it cause any issues? I'd rather we enabled it as it'd aloow the ~45% of use on iOS 10 to scroll this area and we can look at the issue on iOS 11 in a separate branch.

regiskuckaertz: It throws an `undefined` exception, and so everything after it does not run. Yes I think you're right, I'll amend this by testing the value.

regiskuckaertz:
YEAH! Thank you 😄

PR #674, opened 2017-09-27
Add reset styles for buttons

regiskuckaertz:
This is copied verbatim from the reset styles in frontend


regiskuckaertz:
I don't understand why my build is failing 😆

PR #677, opened 2017-10-09
Add hasEpic flag for outbrain  checks

GHaberis:
add placeholder `__HAS_EPIC__` to HTML templates. The native layer should replace with `true` if am epic is to be inserted on the article. Otherwise they should replace it with an empty string or false.

When checking whether to render a compliant or non-compliant outbrain widget we now check to see whether `GU.opts.hasEpic` is `true`. If so render a non-compliant widget.

Why make this change? The current outbrain code checks `prose.lastElementChild.classList.contains('creative-container')` for the presence of an epic. The problem is the epic may not exist in the DOM at the point this check is made because Epics are currently injected in the DOM. We therefore need an indication an Epic will be present in the initial DOM, `__HAS_EPIC__` is that indication.


alfavata:
@GHaberis, could you please update this branch with the fix recently merged in `master`? I'm having troubles testing my PR otherwise.

GHaberis:
@alfavata just updated this branch with `master`.

ArticleTemplates/assets/js/modules/outbrain.js
```diff
@@ -11,7 +11,7 @@ define(function () {
     function isNonCompliant() {
         var prose = document.querySelector('.article__body > div.prose ');
 
-        if (prose && prose.lastElementChild && prose.lastElementChild.classList.contains('creative-container')) {
+        if (prose && prose.lastElementChild && GU.opts.hasEpic) {
```
alfavata: Do you still need to check `prose && prose.lastElementChild`? (or indeed, do you still need `prose` at all)

PR #682, opened 2017-11-22
Translate enum strings to enum numbers

regiskuckaertz:
Off the [fix](https://github.com/guardian/atom-renderer/releases/tag/v0.8.16) we released yesterday, we need to make sure the apps still receive data in the form it is expecting. This should be temporary until both apps are updated to handle strings instead of numbers, but that would delay the release for another few weeks.

It's working as expected:
screen shot 2017-11-22 at 14 19 53

### DIFF
--- a/ArticleTemplates/assets/js/modules/atoms/services.js
+++ b/ArticleTemplates/assets/js/modules/atoms/services.js
@@ -5,6 +5,57 @@ define([
   util,
   viewport
 ) {
+  // Source: https://github.com/guardian/ophan/blob/master/event-model/src/main/thrift/componentevent.thrift
+  var actions = {
+    INSERT: 1,
+    VIEW: 2,
+    EXPAND: 3,
+    LIKE: 4,
+    DISLIKE: 5,
+    SUBSCRIBE: 6,
+    ANSWER: 7,
+    VOTE: 8,
+    CLICK: 9
+  };
+
+  var componentTypes = {
+    READERS_QUESTIONS_ATOM: 1,
+    QANDA_ATOM: 2,
+    PROFILE_ATOM: 3,
+    GUIDE_ATOM: 4,
+    TIMELINE_ATOM: 5,
+    NEWSLETTER_SUBSCRIPTION: 6,
+    SURVEYS_QUESTIONS: 7,
+    ACQUISITIONS_EPIC: 8,
+    ACQUISITIONS_ENGAGEMENT_BANNER: 9,
+    ACQUISITIONS_THANK_YOU_EPIC: 10,
+    ACQUISITIONS_HEADER: 11,
+    ACQUISITIONS_FOOTER: 12,
+    ACQUISITIONS_INTERACTIVE_SLICE: 13,
+    ACQUISITIONS_NUGGET: 14,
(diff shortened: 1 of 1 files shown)


tomrf1:
👍 looks reasonable

regiskuckaertz:
Pending release of [this one](https://github.com/guardian/mobile-apps-api/pull/1022)

ArticleTemplates/assets/js/modules/atoms/services.js
```diff
@@ -5,6 +5,57 @@ define([
...
+    ACQUISITIONS_NUGGET: 14,
+    ACQUISITIONS_STANDFIRST: 15,
+    ACQUISITIONS_THRASHER: 16,
+    ACQUISITIONS_EDITORIAL_LINK: 17,
+    ACQUISITIONS_MANAGE_MY_ACCOUNT: 18,
+    ACQUISITIONS_BUTTON: 19,
+    ACQUISITIONS_OTHER: 20
+  };
+
+  // Temporary fix:
```
alfavata: I'm not sure about the fact that this fix should be temporary... what would be the advantage of having to define the String-to-Int conversion on two different platforms, with all sorts of possible misalignments, rather than keeping it all in one place?

regiskuckaertz: If I'm not mistaken, you are doing the int to string conversion in the app (because Ophan only understand strings), so the work above is doing string to int to make the app happy, but overall both should be unnecessary; i.e. you can take a string in and send it to Ophan as is.

regiskuckaertz: (the work above also doubles the maintenance cost of updating the thrift definitions)

alfavata: That is not correct, I'm afraid :)
Both `Action` and `ComponentType` are enums ([docs here](https://dashboard.ophan.co.uk/docs/thrift/componentevent.html#Enum_Action)), so we send numeric values in the Thrift objects, and Ophan wouldn't accept strings for those fields.

alfavata: (Indeed, the best approach would be to start sending numeric values as far upstream as possible!)

regiskuckaertz: I agree 1000% with you. We should _never_ have use string serialization, this is totally something that the scrooge library allowed and it was a terrible idea. Numbers all along!

The sad fact is that Ophan is expecting strings ☹️ You can see it in action here:
https://github.com/guardian/ophan/blob/master/the-slab/app/extractors/ComponentEventExtractor.scala#L44-L56

regiskuckaertz: (I was initially sending numbers and when we launched in frontend, we noticed nothing was coming through in Ophan. Switching to string solved this, but I agree with you that it was a terrible idea to use strings in Ophan.)

regiskuckaertz: Scratch that, I just saw native submissions were handled differently. Then you're right, we'll have to live with that, hopefully one day web submissions will use Thrift enums correctly 😭

PR #684, opened 2017-11-29
define style for inverting already inverted images

glockett:
Uses Media Query to detect if colours are inverted on iOS. Then inverts the colour of the already inverted image. This is used to make images not inverted when the 'invert' accessibility setting is switched on in iOS.

@GHaberis 

Before:

After:


webb04:
@glockett sorry this never got merged. Simone is working on this right now along with some native changes so will close this branch and try to get that one out soon.

PR #688, opened 2017-12-07
Create new fixtures

duarte:
This creates a new set of fixtures to test the app templates with, using the new garnett CSS files and with classes for garnett pillar and garnett content types


GHaberis:
@duarte do you need to rename `garnett-fonts-ios.css` to `fonts-ios.css` as per our discussion yesterday?

PR #704, opened 2017-12-18
Fix liveblog template

GHaberis:
A few bugs were reported in Liveblogs using Garnett. 

The elements with the classes `.article__body`, `.more`, `.loading` and `.tags` need to sit within the elem `.article` but as part of Garnett redesign they had been accidentally moved outside of this elem, causing the reported bugs. This PR moves them back to the correct position.

The change is in the file`liveblogTemplate.html`, additional changes to the test fixtures are prettifying the HTML.


GHaberis:
The issue was reported by @alfavata

PR #706, opened 2017-12-19
Ad position fix

GHaberis:
Ads on tablet layouts are positioned to the top/right of the article body in the app templates. A placeholder reserves the space for the ad. This placeholder was mispositioned on iPad layouts causing the paragraph text to overlap the placeholder after this commit: https://github.com/guardian/mobile-apps-article-templates/commit/c78bef993d6da842ba90b03022344df9f5b80c0d#diff-6394f01c0d5259cee98d42521387e67d

This PR fixes the position.

**Before...**

screenshot 1

**After...**

picture 80

### DIFF
--- a/ArticleTemplates/assets/scss/garnett-modules/_advert-slot.scss
+++ b/ArticleTemplates/assets/scss/garnett-modules/_advert-slot.scss
@@ -1,0 +1,185 @@
+// /*doc
+// ---
+// title: Advert slot
+// name: advert-slot
+// category: Modules
+// ---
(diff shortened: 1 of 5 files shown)


GHaberis:
As reported by @alfavata

GHaberis:
@NataliaLKB thanks for checking it out. The change which caused pre-garnett templates to break was required for a post-garnett world, I've therefore created a copy of `_advert-slot.scss` and put it here, `garnett-modules/_advert-slot.scss`, this is only used by garnett articles.

NataliaLKB:
Thanks @GHaberis !

PR #707, opened 2017-12-20
update fonts to use correct headline font filename

GHaberis:
We're no longer using the `agaate` font files. This PR update the app's templates to use the correct file names.

### DIFF
--- a/ArticleTemplates/assets/scss/base/_typography.scss
+++ b/ArticleTemplates/assets/scss/base/_typography.scss
@@ -7,84 +7,84 @@ $fontMap: (
         ios-src: 'url("../fonts/icons.otf")'
     ),
 
-    // Agaate
-    AS31GuardianAgaateSans1WebLight: (
-        family: 'Guardian Agaate Sans Web',
+    // Headline
+    GHGuardianHeadlineLight: (
+        family: 'Guardian Headline Web',
         weight: 200,
-        android-src: 'url("../garnett-fonts/AS31GuardianAgaateSans1Web-Light.ttf")',
-        ios-src: 'url("../garnett-fonts/AS31GuardianAgaateSans1Web-Light.ttf")'
+        android-src: 'url("../garnett-fonts/GHGuardianHeadline-Light.ttf")',
+        ios-src: 'url("../garnett-fonts/GHGuardianHeadline-Light.ttf")'
     ),
-    AS31GuardianAgaateSans1WebLightItalic: (
-        family: 'Guardian Agaate Sans Web',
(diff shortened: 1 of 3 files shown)


GHaberis:
@alfavata @mohammad-haque this is the PR that corresponds to https://github.com/guardian/ios-live/pull/3605.

@jorgeazevedo can you confirm these file paths look ok?

mohammad-haque:
@GHaberis Can you please also rename `garnett-fonts` to `fonts`? This make things simpler in apps.

alfavata:
I wonder why the fonts have "Web" at the end of their family name, if they're the same we're using in the apps' native elements 🤔 
...and consequently I wonder if we _should_ use the same fonts (the `noalts_not_hinted` set) for the apps' native elements and the templates. @paperboyo help 😅 

(also, Jorge is back at work in January)

alfavata:
The  guardian/ios-live#3605 PR also changed `garnett-fonts` to just `fonts`, so that's required for everything to work once both are merged

paperboyo:
@alfavata `Web` added to the font name means it’s coming from the dotcom set (old fonts). There is no `Web` in the name in any version of the GH font anywhere. So I would remove `Web` from anything GH-related. Even more so, because apps have their own version of GH fonts (TTF, non-obfuscated) and shouldn’t be using GH fonts from dotcom (which have their own TTFs but obfuscated).

GHaberis:
@alfavata @mohammad-haque I've updated the file paths and family name as suggested

alfavata:
Great, thanks @paperboyo and @GHaberis 

Just to be clear: besides the name, is it correct that apps templates are using the `noalts_not_hinted` non-obfuscated set?

paperboyo:
> is it correct that apps templates are using the noalts_not_hinted non-obfuscated set

Yes. This way they are 100% dotcom-parity. `Full` have the same glyphset, but contain some OpenType features that a) we aren’t using b) were removed to save weight.

One day the features will come back (not before Composer and stuff will be ready).

alfavata:
Okay, then let's merge both this and guardian/ios-live#3605  :)

GHaberis:
@mohammad-haque @alfavata before I merge can I just confirm the changes @jorgeazevedo made to the build scripts for iOS/Android now copy these fonts into the template's `fonts` folder and not `garnett-fonts` anymore?

alfavata:
My PR updates thats script for iOS. I don't know how that works on Android, to be honest

mohammad-haque:
It's fine for android too.

PR #708, opened 2017-12-21
Use correct Garnett class names

GHaberis:
Previously all the HTML templates contained the following classes on the `body`: `type-__GARNETT_TYPE__` and `pillar-__GARNETT_PILLAR__`. Once the placeholders have been replaced these would result in `type-article` and `pillar-news` for example.

The css however was targetting `garnett--type-article` and `garnett--pillar-news`.

I've therefore updated the HTML templates to use `garnett--type-__GARNETT_TYPE__` and `garnett--pillar-__GARNETT_PILLAR__`, once the placeholders have been replaced these would result in the correct classnames.


GHaberis:
@NataliaLKB thanks for reviewing so quickly and yes that's a good suggestion!

GHaberis:
@alfavata @mohammad-haque this PR means the Garnett styles @duarte has been working on should now be available to view on Garnett enabled builds of your apps.

alfavata:
I thought they already were 😅

PR #709, opened 2018-01-04
Garnett - Replace icons

GHaberis:
Garnett - Replace references to old icons in `.scss` to new icons.


GHaberis:
@paperboyo

paperboyo:
Would be nice to check the rendering as the replacement is larger. But the thing replaced is wrong, so we have to change it anyway...

PR #735, opened 2018-01-17
Updates icon font file to new version

duarte:
Updates icon font file to new version, now that we can have it in the repository. Unless it doesn't make sense to have it here @jorgeazevedo / @GHaberis ?

Also, we need to update the fonts in the app overall soon in a separate action.


alfavata:
That font is being replaced by our version when the app is build...

duarte:
So it's only used for local development — I will merge

PR #741, opened 2018-02-02
Handle Horizontal Swipe on Audio slider

GHaberis:
On Android the template needs to inform the native layer that a horizontal swipe has begun/ended, the Android layer will ignore horizontal swiping when this occurs, meaning users can use the slider on Audio templates. `GuardianJSInterface.registerRelatedCardsTouch` is the interface the JS can call to do this.


GHaberis:
This fixes an issue reported by @jorgeazevedo

jorgeazevedo:
Yeah I tested this branch on Android and it fixes the issue 👍

PR #751, opened 2018-03-06
Removed negative margin & updated fixture

zeftilldeath:
Multi-rule was overlapping meta on Guardian view template.

# before
screen shot 2018-03-06 at 15 50 20

# after
screen shot 2018-03-06 at 15 49 39


alfavata:
(Now that's taking "reviewing a PR" to the next level!)

PR #755, opened 2018-03-12
Immersive headline

duarte:
Updates the immersive headline design.

**Before**

**After**

### DIFF
--- a/ArticleTemplates/articleTemplateContainerImmersive.html
+++ b/ArticleTemplates/articleTemplateContainerImmersive.html
@@ -14,7 +14,7 @@
             </h1>
         </div>
 
-        <div class="standfirst standfirst">
+        <div class="standfirst">
             <div class="standfirst__text">__STANDFIRST__</div>
             __IMMERSIVE_BYLINE__
         </div>
--- a/ArticleTemplates/articleTemplateContainerImmersive.html
+++ b/ArticleTemplates/articleTemplateContainerImmersive.html
@@ -12,14 +12,9 @@
             <h1 class="article__headline headline selectable">
                 __TITLE__
             </h1>
-
-            <div class="standfirst standfirst--tablet">
-                <div class="standfirst__text">__STANDFIRST__</div>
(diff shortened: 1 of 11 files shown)


ArticleTemplates/articleTemplateContainerImmersive.html
```diff
@@ -5,20 +5,16 @@
...
 
-        <div class="standfirst standfirst--mobile">
+        <div class="article__header-info">
+            __IMMERSIVE_SERIES__
+            <h1 class="article__headline headline selectable">
+                __TITLE__
+            </h1>
+        </div>
+
+        <div class="standfirst standfirst">
```
GHaberis: double use of `standfirst` class here 👯

ArticleTemplates/assets/scss/garnett-type/_immersive.scss
```diff
@@ -101,70 +100,83 @@
...
+                margin-left: auto;
+                margin-right: auto;
                 padding: {
-                    left: 0;
-                    right: 0;
+                    left: base-px(1);
+                    right: 340px;
+                }
+                width: 1200px;
+                transform: translateX(228px); //240 - 8
```
GHaberis: 🕵️  not quite sure what the numbers represent, but 240 - 8 is `232px`. Could it be worth assigning `228` to a variable as it's referenced in a few places which I imagine would all need updating if it were to change?

duarte: 🤣I dreamt of an 8 pt gutter on this template but immediately regretted it haha
Will update everything, thanks george!

duarte: I've updated it to use the SASS calculation `240px - base-px(1)` instead, which I think is easier to understand — does that seem good to you?

GHaberis: Sounds good to me

duarte:
Thanks for reviewing George!

PR #756, opened 2018-03-12
Add templates for callout campaigns

regiskuckaertz:
We're adding the ability for journalists in the participation team to create "Take part" campaigns, where readers can submit information that will be used in the news process.

The data is already available in the `campaigns` field of the MAPI payload, all we need is pick a "take part" campaign if there is one at all, and feed it into the templates defined in this PR.

The main template is `calloutTemplate`, defining the root of our component. It is inserted into the body of an article right after the `__BODY__`. The heading is made of a title and an optional description (`calloutTemplateDescription`). The content of a campaign is a form to submit data, and a form is a sequence of fields that may either be a textarea (`calloutTemplateText`), a file upload (`calloutTemplateFile`), a list of checkbox/radio input (`calloutTemplateMulti`, `calloutTemplateCheckbox` and `calloutTemplateRadio`) or an abstract input (`calloutTemplateInput`).

Currently waiting on a design for these so this PR should not be merged before the following boxes have been ticked off:

- [x] Add the CSS
- [x] Test locally
- [ ] Test on devices


ArticleTemplates/calloutTemplate.html
```diff
@@ -0,0 +1,11 @@
+<details class="article_callout">
+  <summary class="article_callout_heading">
+    <div class="article_callout_title">
```
hosamaly: Out of curiosity, is there a reason to prefer a class name like `article_callout_title` to multiple classes, like `article callout title`?

regiskuckaertz: These would lead to two opposite outcomes. My version essentially reduces to a boolean AND: `∀x. article(x) ∧ callout(x) ∧ title(x)`, whereas yours reduces to an OR: `∀x. article(x) ∨ callout(x) ∨ title(x)`. So in the latter case, your element would inherit all the styles from either `article`, `callout` or `title`.

SLambrakis:
iOS: [GLA-500](https://theguardian.atlassian.net/browse/GLA-500)
Android: [AND-12069](https://theguardian.atlassian.net/browse/AND-12069)

ArticleTemplates/calloutTemplate.html
```diff
@@ -0,0 +1,11 @@
...
+      __CALLOUT_TITLE__
+    </div>
+    __CALLOUT_DESCRIPTION__
+  </summary>
+  <form method="post" action="x-gu://participate/__CALLOUT_ID__" class="article_callout_body">
+    __CALLOUT_BODY
```
GHaberis: Are you missing the trailing `__` on `__CALLOUT_BODY`?

regiskuckaertz: Oh yeah I am, thanks!

regiskuckaertz: ... I'm wondering whether the form templates should be called `formXYZ` instead, their use in callouts is only circumstantial. What do you think?

GHaberis:
One thing, inserting an element beneath the article body will mean Outbrain is non-compliant, we'll need to account for this in the 'isNonCompliant' check [here](https://github.com/guardian/mobile-apps-article-templates/blob/008ae5544631d1514626f612e62c5e611da7ff3a/ArticleTemplates/assets/js/modules/outbrain.js#L12 ) by checking the DOM for existence of a `.article_callout` element.

regiskuckaertz:
@GHaberis yeah that's what I thought, will do that 👍

ArticleTemplates/calloutTemplate.html
GHaberis: Good idea, they look useful, giving them more generic names leaves the possibility for reusing them elsewhere without having to worry about renaming them (and all the native work this'd entail) or using "callout" templates in a non-callout context.

regiskuckaertz:
@GHaberis is it possible to get the new typeface in this project? would be quite handy for me to work out the correct weights and show to designers.

GHaberis:
@regiskuckaertz sorry for delay, I've been away on leave. The new typeface could and should now be added to this project. I'll arrange this and inform you when it's ready.

GHaberis:
I've just checked and these fonts were added to the repo for the redesign:

- GHGuardianHeadline-Black.ttf
- GHGuardianHeadline-Bold.ttf
- GHGuardianHeadline-Light.ttf
- GHGuardianHeadline-LightItalic.ttf
- GHGuardianHeadline-Medium.ttf
- GHGuardianHeadline-Regular.ttf
- GHGuardianHeadline-Semibold.ttf

Are you expecting another typeface to be available @regiskuckaertz?

regiskuckaertz:
You're right, I forgot that variable substitution is occurring inside the apps. Will add default values somewhere. Thanks!

regiskuckaertz:
Closing as not relevant as per discussion.

PR #757, opened 2018-03-13
Fix yarn lock

SiAdcock:
`yarn.lock` appears to be out of date with the dependencies in this project. Maybe I'm going crazy.

I ran:

```
$ rm -rf node_modules
$ yarn
```

And the `yarn.lock` was updated to the tune of the changes in this pull request. Let me know if I've done something wrong, but I'm inclined to believe we need to merge these changes to keep Yarn deterministic.

I'm using Yarn v1.3.2, but this also happens in the latest (v1.5.1)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -17,6 +17,13 @@ accepts@1.3.3:
     mime-types "~2.1.11"
     negotiator "0.6.1"
 
+accepts@~1.3.4:
+  version "1.3.5"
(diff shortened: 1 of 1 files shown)


GHaberis:
hmm, could dependencies have been added to`package.json` and installed via `npm`, meaning `yarn.lock` didn't pickup these updates? possibly, anyway I'm inclined to agree that this PR is necessary!

SiAdcock:
Yeah, it looks like it's dependencies of `grunt-contrib-connect`

PR #759, opened 2018-03-13
Remove dev server

SiAdcock:
The dev server is not currently used. 

picture 513

This change removes the dev server, its index page and its dependencies.


GHaberis:
This devSever was setup during Garnett by @duarte in this PR https://github.com/guardian/mobile-apps-article-templates/pull/687.

The PR above means the route `http://localhost:8000/test/garnett-fixture/article-news.html` would work with reloads on changes.

The issue we have is the `index.html` page rendered at `http://localhost:8000` is completely unrelated and out of date. We could install [grunt-connect-route](https://www.npmjs.com/package/grunt-connect-route) and create a route for `http://localhost:8000` that does something a little more useful.

SiAdcock:
Ah cool, nice one! Some docs would be useful too!

PR #760, opened 2018-03-13
Inline links in standfirst

SiAdcock:
Currently, the last link in the standfirst is set to `inline-block`. This pushes proceeding text down onto this next line

This change makes the standfirst inline again.

## Screenshots

**before**

screenshot 2

**after**

picture 514

## Discussion

I believe the standfirst was being set to `inline-block` in order to add bottom margin if and only if _the link is at the very end of the standfirst_. However, this selector is targeting _the last link in the standfirst_, regardless of its location within the standfirst.

I'm not sure if there is a selector that can achieve the original purpose. @duarte let me know if you are happy to lose this bottom margin, or if we need to come up with an alternative.

**With margin (before this change)**

picture 516

**Without margin (after this change)**

picture 515

### DIFF
--- a/ArticleTemplates/assets/scss/garnett-modules/content/_standfirst.scss
+++ b/ArticleTemplates/assets/scss/garnett-modules/content/_standfirst.scss
@@ -24,9 +24,4 @@
     b {
         font-weight: 700;
     }
-
-    p:last-child a:last-child {
-        display: inline-block;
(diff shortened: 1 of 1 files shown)


duarte:
Thanks for doing this Simon!
You're right about the intention of that selector, but it was a terrible implementation and it's good to remove it! :)

PR #761, opened 2018-03-14
Add new templates for collections (AA-14)

maxspencer:
This PR adds two new templates the apps can use to render a new type of content which we're calling a "collection". This will eventually look something like the design below provided by @blongden73. The image at the top and title are native, everything below is rendered using these HTML templates.

Not shown below are the "Read full article ->" links that should appear at the end of each "block", I have put those in the template though.

`collectionsTemplate.html` is a copy of `interactiveTemplate.html` but with the `interactives.css` swapped for the CSS used in `articleTemplate.html`.

I will need to add to the CSS later, but I wanted to open this PR now to try and get feedback on the general direction, if possible?

collection_expanded

### DIFF
--- a/ArticleTemplates/collectionBlockTemplate.html
+++ b/ArticleTemplates/collectionBlockTemplate.html
@@ -1,10 +1,10 @@
-<div class="collection-block-container" style="border: solid 1px #ccc; padding: 10px; margin: 10px;">
-    <h2>
-        <a href="__BLOCK_LINK__"><span class="collection-sub-item-title-line-1">__BLOCK_TITLE_1__</span><br/>
-        <span class="collection-sub-item-title-line-2">__BLOCK_TITLE_2__</span></a>
-    </h2>
+<div class="collection-block">
+    <h3 class="collection-block-title">
+        <a href="__BLOCK_LINK__"><span class="collection-sub-item-title-line-1" style="color: __BLOCK_TITLE_1_COLOUR__;">__BLOCK_TITLE_1__</span>
+        <span class="collection-sub-item-title-line-2" style="color: __BLOCK_TITLE_2_COLOUR__;">__BLOCK_TITLE_2__</span></a>
+    </h3>
     __BLOCK_BODY__
-    <p>
+    <p class="collection-read-full-container">
         <a class="collection-read-full-link" href="__BLOCK_LINK__">Read full article</a>
     </p>
 </div>
--- a/ArticleTemplates/collectionBlockTemplate.html
+++ b/ArticleTemplates/collectionBlockTemplate.html
@@ -1,0 +1,10 @@
+<div class="collection-block-container" style="border: solid 1px #ccc; padding: 10px; margin: 10px;">
+    <h2>
+        <a href="__BLOCK_LINK__"><span class="collection-sub-item-title-line-1">__BLOCK_TITLE_1__</span><br/>
+        <span class="collection-sub-item-title-line-2">__BLOCK_TITLE_2__</span></a>
+    </h2>
+    __BLOCK_BODY__
+    <p>
+        <a class="collection-read-full-link" href="__BLOCK_LINK__">Read full article</a>
+    </p>
+</div>
--- a/ArticleTemplates/collectionBlockTemplate.html
+++ b/ArticleTemplates/collectionBlockTemplate.html
@@ -5,6 +5,6 @@
     </h3>
     __BLOCK_BODY__
     <p class="collection-read-full-container">
-        <a class="collection-read-full-link" href="__BLOCK_LINK__">Read full article</a>
+        <a class="collection-read-full-link" href="__BLOCK_LINK__">Read full article <span data-icon="→" aria-hidden="true"></span></a>
     </p>
 </div>
--- a/ArticleTemplates/collectionTemplate.html
+++ b/ArticleTemplates/collectionTemplate.html
@@ -1,11 +1,10 @@
 <!DOCTYPE html>
 <html lang="en-US">
 <head>
-    <title>Interactive</title>
+    <title>Collection</title>
     <meta charset="UTF-8" />
     <meta name="viewport" content="initial-scale=1, maximum-scale=1" />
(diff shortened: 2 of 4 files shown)


ArticleTemplates/collectionTemplate.html
```diff
@@ -0,0 +1,45 @@
...
+
+    __BODY__
+
+    <script type="text/javascript">
+        GU.bootstrap.init({             
+            asyncStylesFilename: "__ASYNC_STYLES__",
+            fontSize: "__FONT_SIZE__",
+            platform: "__PLATFORM__",
+            isOffline: "__IS_OFFLINE__" ? true : false,
+            contentType: "interactive",
```
GHaberis: the value of `contentType` will need updating to something appropriate for this article type

ArticleTemplates/collectionTemplate.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
+<html lang="en-US">
+<head>
+    <title>Interactive</title>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="initial-scale=1, maximum-scale=1" />
+    <link rel="stylesheet" type="text/css" media="all" id="fontStyles" href="__TEMPLATES_DIRECTORY__assets/css/fonts-__PLATFORM__.css" />
+    <!--<link rel="stylesheet" type="text/css" media="all" href="__TEMPLATES_DIRECTORY__assets/css/interactive.css" />-->
```
GHaberis: `interactive.css` exists because the interactive template has a very minimal set of styles applied to it.

All other templates use a combination of `__SYNC_STYLES__.css` and `fonts-__PLATFORM__.css` in the `<head>` and `__ASYNC_STYLES__` which are loaded by the JS. This template should be no different, so I'd just remove the commented out `link`.

ArticleTemplates/collectionBlockTemplate.html
```diff
@@ -0,0 +1,10 @@
+<div class="collection-block-container" style="border: solid 1px #ccc; padding: 10px; margin: 10px;">
```
GHaberis: These inline styles should be removed and out in the css.

ArticleTemplates/collectionTemplate.html
```diff
@@ -0,0 +1,45 @@
...
+    <link rel="stylesheet" type="text/css" media="all" href="__TEMPLATES_DIRECTORY__assets/css/__SYNC_STYLES__.css" />
+    <script type="text/javascript">
+        window.GU = window.GU || {};
+    </script>
+    <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/polyfills/promise.js"></script>
+    <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/polyfills/requestAnimationFrame.js"></script>
+    <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/bootstrap.js"></script>
+</head>
+
+<body class="__FONT_SIZE__ __PLATFORM__ __IS_OFFLINE__ __DISPLAY_HINT__" data-fullscreen-interactive data-content-type="interactive" data-ads-enabled="__ADS_ENABLED__" data-ads-enable-hiding="__ADS_ENABLE_HIDING__" data-ads-config="__ADS_CONFIG__" style="margin-top: __ACTIONBARHEIGHT__px;" data-font-size="__FONT_SIZE_INT__" data-page-id="__PAGE_ID__" data-template-directory="__TEMPLATES_DIRECTORY__" data-mpu-after-paragraphs="__MPU_AFTER_PARAGRAPHS__" data-use-ads-ready="__ADS_FAST_CALLBACK__">
```
GHaberis: the the value of `data-content-type` will need updating to something appropriate for this article type

ArticleTemplates/collectionTemplate.html
alfavata: We were talking about this with @chriswilk and @alexduf: if we had a new `collection` value for `contentType`, old iOS clients would filter out collections automatically.

maxspencer:
I'm happy with this now (for a hacky first version). If anyone can give me a +1 I'd be much obliged.

ArticleTemplates/collectionTemplate.html
```diff
@@ -0,0 +1,50 @@
...
+        window.GU = window.GU || {};
+    </script>
+    <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/polyfills/promise.js"></script>
+    <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/polyfills/requestAnimationFrame.js"></script>
+    <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/bootstrap.js"></script>
+</head>
+
+<body class="garnett--type-collection garnett--pillar-__GARNETT_PILLAR__ tone--__SECTION_TONE__ __FONT_SIZE__ __PLATFORM__ __IS_OFFLINE__ advert-config--__ADS_CONFIG__ __IS_ADVERTISING__ __DISPLAY_HINT__" data-content-type="article" data-ads-enabled="__ADS_ENABLED__" data-ads-enable-hiding="__ADS_ENABLE_HIDING__" data-ads-config="__ADS_CONFIG__" style="margin-top: __ACTIONBARHEIGHT__px;" data-font-size="__FONT_SIZE_INT__" data-template-directory="__TEMPLATES_DIRECTORY__" data-page-id="__PAGE_ID__" data-mpu-after-paragraphs="__MPU_AFTER_PARAGRAPHS__" data-use-ads-ready="__ADS_FAST_CALLBACK__" data-content-section="__SECTION__">
+
+    <div class="article article--feature" id="feature-article-container">
```
GHaberis: Can you remove the class `article--feature` and the `id`? They may cause issues with `feature` style rules being applied?

maxspencer:
Thanks @GHaberis, I've replaced the `feature` styles with `standard` ones. Are you happy for me to merge this now?

maxspencer:
I've got an informal, out-of-band, thumbs up for this so I'm going to merge it.

PR #768, opened 2018-03-28
Colour sorting

duarte:
No visual changes (🤞) just better organisation of the pillar colours file.

This stops using local variables in each pillar file, moving pillar colours to the global colour variables file. Now pillar colours can be used anywhere! Also moves the code to make media backgrounds black in all pillars from pillar colour file to the media-all file.


duarte:
Thanks for the quick review Simon! :)

PR #774, opened 2018-04-13
Underline links in liveblog standfirsts

duarte:
Links were not getting underlined in liveblog standfirsts. This PR also adopts article-style typography for liveblog standfirsts — bold, smaller text.

**Before**

**After**


duarte:
Thanks George!

PR #776, opened 2018-04-24
Add atoms to football and cricket articles

regiskuckaertz:


alfavata:
Thanks again @regiskuckaertz -- I'll leave to @GHaberis to actually review the PR though :)

PR #778, opened 2018-04-24
Tags in special reports

duarte:
With the special report page background becoming a little darker, tag button backgrounds became hard to see. This restores their visibility.

This PR also removes some tag-specific colour vairables that shouldn't be necessary anymore


alfavata:
Please merge this before Friday! :)

PR #780, opened 2018-05-02
adds a collection fixture

duarte:
This just adds a collection fixture so we can update the collection design more easily.

I think I've made the right changes to use this HTML as a fixture. @GHaberis can you please check this is OK?

### DIFF
--- a/test/garnett-fixture/collection.html
+++ b/test/garnett-fixture/collection.html
@@ -1,0 +1,122 @@
+<html lang="en-US"><head>
+    <title>Collection</title>
+    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


GHaberis:
Everything looks OK. The only thing i'd say is...I don't know what a `collection` looks like but it looks like this fixture was originally a `deadBlog`, is that ok? Would a standard article be a closer match?

duarte:
The collection page summarises (and links to) a few other articles, and I think each article it links to is built similarly to a liveblog block (there's a screenshot in #761).

This is just to save the output from the app tho. We could merge this and then update the fixture if the template changes to be based on the standard article?

PR #788, opened 2018-06-05
Speed up unit tests by using fake timers

GHaberis:
The unit tests previously used `setTimeout` to test asynchronous code, this meant the tests were sloooooooow. I've replaced all setTimeouts and now use `sinon`'s implementation of `useFakeTimers`. The tests are much faster now , previously they took 15+ seconds 😖 now they take 2 seconds :boom:

### DIFF
--- a/test/spec/unit/modules/youtubeTest.js
+++ b/test/spec/unit/modules/youtubeTest.js
@@ -10,10 +10,11 @@ define([
      * causing tests to intermittently fail on CircleCI.
      * Skipping for now, but we should revisit.
      */
-    describe.skip('ArticleTemplates/assets/js/modules/youtube', function () {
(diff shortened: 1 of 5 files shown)


test/spec/unit/modules/youtubeTest.js
```diff
@@ -10,10 +10,11 @@ define([
      * causing tests to intermittently fail on CircleCI.
      * Skipping for now, but we should revisit.
      */
-    describe.skip('ArticleTemplates/assets/js/modules/youtube', function () {
-        this.timeout(15000);
+    describe('ArticleTemplates/assets/js/modules/youtube', function () {
+        // this.timeout(15000);
```
SiAdcock: I guess we can kill this comment

PR #789, opened 2018-06-05
Ao related items

aodhol:
This PR:

- Removes some of the legacy Related Items related HTML / JS.
- Remove an unused test and dependency.
- Adds two functions to size and position a container where the 'native' Related Items component will be positioned.
- Adds a [Mutation Observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) on the body to listen for changes before positioning the container.


GHaberis:
Hey Aodh, this is failing the tests on CircleCI because there were some tests for `cards.js`. We can either remove that test spec, or if you fancy it write some tests to replace the old ones.

There's also a dependency `flipSnap` that was only used by the old implementation of `cards.js`. We can now remove this from `package.json` and there are a few other locations we'll need to remove references to it.

ArticleTemplates/assets/js/bootstrap.js
```diff
@@ -34,7 +32,9 @@
             'initMpuPoller',
             'videoPositioning',
             'getArticleHeight',
-            'injectInlineArticleMembershipCreative'
+            'injectInlineArticleMembershipCreative',
+            'getRelatedContentPosition',
+            'setRelatedItemsHeight'
```
GHaberis: Minor thing but maybe we can rename `setRelatedItemsHeight` to `setRelatedContentHeight` as it'll help show it's related to the "related content" container we refer to elsewhere in the code.

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,39 @@
...
-    function ready() {
-        if (!initialised) {
-            initialised = true;
-            setupGlobals();
-        }
+    function setupGlobals() {
+        window.getRelatedContentPosition = getRelatedContentPosition;
+        window.applyNativeFunctionCall("getRelatedContentPosition");
+        window.setRelatedItemsHeight = setRelatedItemsHeight;
+        window.applyNativeFunctionCall("setRelatedItemsHeight");
```
GHaberis: If you do rename `setRelatedItemsHeight` to `setRelatedContentHeight` it'll need updating here as well.

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,45 @@
...
-
-        if (relatedContentWrapper) {
-            if (!html) {
-                articleCardsFailed();
-            } else {
-                relatedContentWrapper.innerHTML = html;
-                snapToGrid();
-            }
+    function ready(config) {
+        if (!initialised) {
```
GHaberis: As this only needs to run on iOS could you also add `&& GU.opts.platform === 'ios'` to this conditional?

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,45 @@
...
-        if (!initialised) {
-            initialised = true;
-            setupGlobals();
-        }
+    function setupGlobals() {
+        window.getRelatedContentPosition = getRelatedContentPosition;
+        window.applyNativeFunctionCall("getRelatedContentPosition");
+        window.setRelatedContentHeight = setRelatedContentHeight;
+        window.applyNativeFunctionCall("setRelatedContentHeight");
+        observer = new window.MutationObserver(function(mutations) {
```
GHaberis: Can this `observer` logic be moved out of `setupGlobals` into `ready`. The convention is `setupGlobals` just exposes methods on the `window` for the native layer to call.

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,45 @@
...
+            window.webkit.messageHandlers.bodyMutationNotification.postMessage({body: {} });
```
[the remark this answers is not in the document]
GHaberis: I see you're posting `{body: {} }` to `bodyMutationNotification` on every mutation? On `bodyMutationNotification` does the iOS layer request the position of the related content? 

If so, could you actually run the check on the position of the related content container in this callback? And if it has changed at all then post that data to `bodyMutationNotification`. That'd save some back and forth between the native and web layers?

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,45 @@
...
-            window.addEventListener('resize', util.debounce(onResize.bind(null, relatedContentList), 100));
-        }
-    }
-
-    function onResize(relatedContentList) {
-        if (flipSnap && flipSnap.destroy && relatedContentList) {
-            flipSnap.destroy();
-            relatedContentList.removeAttribute('style');
-            setUpFlipSnap(relatedContentList);
+            observer.observe(relatedContent, { attributes: true });
```
GHaberis: Do we need to set up an new `observer.observe` everytime here? Is setting it up once on the body `observer.observe(window.document.body, { ... })` not enough?

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,45 @@
...
-                relatedContentList.addEventListener('touchstart', onTouchStart);
-                relatedContentList.addEventListener('touchend', onTouchEnd);
-            }
-        }            
-    }
-
-    function onTouchStart() {
-        window.GuardianJSInterface.registerRelatedCardsTouch(true);
+    function setRelatedContentHeight(height) {
+        document.querySelector('.related-content').style.height = height + 'px';
```
GHaberis: Can we assign this `querySelector` to a variable `relatedContent` and  wrap this in a conditional... 

`if (relatedContent) { relatedContent.style.height = height + 'px'; }`

ArticleTemplates/assets/js/modules/cards.js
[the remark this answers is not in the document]
aodhol: Great idea!

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,47 @@
...
-            if (!html) {
-                articleCardsFailed();
-            } else {
-                relatedContentWrapper.innerHTML = html;
-                snapToGrid();
-            }
+    function ready(config) {
+        if (!initialised && GU.opts.platform === 'ios') {
+            initialised = true;
+            new window.MutationObserver(function(mutations) {
```
GHaberis: I was thinking we could assign the last result of `getRelatedContentPosition()` sent to the native layer top a variable, then in the `function(mutations){ ... }` after we call `getRelatedContentPosition()` we can compare with the last result stored in memory, only if there's a difference do pass the result to `window.webkit.messageHandlers.bodyMutationNotification.postMessage`. This could save us sending position data to the native layer even if the position of the related content container hasn't changed. What do you think? Would it save the native layer continuously repositioning the native related cards?

ArticleTemplates/assets/js/modules/cards.js
```diff
@@ -1,96 +1,52 @@
...
+                var newRelatedContentPosition = getRelatedContentPosition();
+                if (newRelatedContentPosition !== existingRelatedContentPosition) {
+                    existingRelatedContentPosition = newRelatedContentPosition;
```
GHaberis: OK this comparison will always return false as you can't test for (deep) equality between to objects like this. You have 2 options here. Iterate over each property of the 2 objects to see if there are any differences. Or stringifying both objects will work in this case: `JSON.stringify(newRelatedContentPosition) !== JSON.stringify(existingRelatedContentPosition)`

PR #791, opened 2018-06-06
fix element-interactive embed bug

GHaberis:
`.element-interactive` emdeds in articles were wider than the width of the viewport on larger devices (eg. iPad), resulting in the lefthand side of the immersive being partially hidden.

An example can be seen in the bug report: https://trello.com/c/R11K2GIS/28-apps-garnett-related-article-embed-bug

This was caused by a `width` being set on these elements in `_article--shared.scss`. Removing this width means the`.element-interactive` inherits the style `width:auto`.

Fixed example...

screen shot 2018-06-06 at 10 45 15

### DIFF
--- a/ArticleTemplates/assets/scss/garnett-layout/_article--shared.scss
+++ b/ArticleTemplates/assets/scss/garnett-layout/_article--shared.scss
@@ -41,14 +41,6 @@
         // Interactives within article body
         .element-interactive {
             width: auto;
-
-            @include mq(col2, col3) {
-                width: calc(#{(100% / (100% - cols($base-3, 3))) * 100%} - #{base-px(1)});
-            }
-
-            @include mq($from: col3) {
-                width: calc(#{(100% / (100% - cols($base-4, 4))) * 100%} - #{base-px(1)});
-            }
         }
 
         .element-video:not(.element-youtube) {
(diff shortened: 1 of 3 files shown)


ArticleTemplates/assets/scss/garnett-layout/_article--shared.scss
```diff
@@ -41,14 +41,6 @@
...
             width: auto;
-
-            @include mq(col2, col3) {
-                width: calc(#{(100% / (100% - cols($base-3, 3))) * 100%} - #{base-px(1)});
-            }
-
-            @include mq($from: col3) {
-                width: calc(#{(100% / (100% - cols($base-4, 4))) * 100%} - #{base-px(1)});
-            }
         }
```
GHaberis: @duarte these style rules were copied over from pre-garnett world. They look like they're trying to do something very clever, I'm not really sure what to be honest, they however caused this bug so I've removed them. Let me know if you think there's an alternative.

alfavata:
Yay 🎉 thanks @GHaberis, much appreciated!

ArticleTemplates/assets/scss/garnett-layout/_article--shared.scss
duarte: I think this was broken by a garnett PR I did where I changed the text column width on tablet. It's possible that if we update the cols() parameters to match the ones on lines 10 and 14 on _article--shared.scss, it would restore the full-width functionality.

https://github.com/guardian/mobile-apps-article-templates/blob/1d43f00676a64dcb43e9ab7249de39d4aec26781/ArticleTemplates/assets/scss/garnett-layout/_article--shared.scss#L10

PR #792, opened 2018-06-06
Render participation campaigns

regiskuckaertz:
See also https://github.com/guardian/mobile-apps-api/pull/1154

Adds styling and behaviour for participation campaigns. Mostly copy-pasted from Web, except for the JS where I'm not turning the form data into a JSON object: I'm assuming the client shell can take care of that.

TODO
- ~~[ ] Apps team to provide a form data processor~~
- [x] Test rendering with related work in MAPI


regiskuckaertz:
Yikes:

screen shot 2018-06-06 at 14 38 47

and

screen shot 2018-06-06 at 14 40 40

ArticleTemplates/assets/js/bootstraps/campaign.js
```diff
@@ -0,0 +1,35 @@
...
+      if (campaign) {
+        initCampaign(campaign);
+      }
+    }
+
+    function initCampaign(campaign) {
+      campaign.onSubmit(function () {
+        const data = new FormData(campaign);
+        // TODO form submission URL
+        const request = new XMLHttpRequest('POST', '#');
```
regiskuckaertz: @alfavata so you're aware, this is the place where I'll need to know where to send the data back to the client app and in which format (currently in `multipart/form-data`)

regiskuckaertz: e.g. `new XMLHttpRequest('POST', 'x-gu://callout')`

alfavata: @aodhol can explain how that won't work 😅

aodhol: @regiskuckaertz XHR won't work because Web / native communication doesn't function as client / server rather the native layer intercepts _location_ changes and parses the URL. I'd suggest using JavaScript to gather the data from the various fields constructing an object that is then JSON stringified, URL encoded appended to the URL for us to then parse in the native layer..

regiskuckaertz: Yep!

ArticleTemplates/assets/js/bootstraps/campaign.js
```diff
@@ -11,11 +11,11 @@ define([], function () {
...
-        // TODO form submission URL
-        const request = new XMLHttpRequest('POST', '#');
-        request.onload(displayConfirmation.bind(null, campaign));
-        request.onerror(displayError.bind(null, campaign));
-        request.send(data);
+        const json = {};
+        for([key, value] of data.entries) {
+          json[key] = value;
+        }
+        location.hash = JSON.stringify(json);
```
regiskuckaertz: @aodhol like this?

alfavata: Could it be the location's query rather than hash?

ArticleTemplates/assets/js/bootstraps/campaign.js
GHaberis: @regiskuckaertz once the native layer have exposed an `x-gu://callout` interface for you to pass data to you can use `signalDevice` exposed by `util.js`, this is used elsewhere in the code for passing data to the native layer. 

@aodhol would it be possible for the native layer to expose a JS function on the window now rather than using the `x-gu://callout` method?

ArticleTemplates/assets/js/bootstraps/campaign.js
regiskuckaertz: Whatever works 🤗  as long as it works on both platforms and the app layer catches the call without reloading the page, I'm all good.

ArticleTemplates/assets/js/bootstraps/campaign.js
regiskuckaertz: Whatever works 🤗  as long as it works on both platforms and the app layer catches the call without reloading the page, I'm all good. The only important thing is that I can register callbacks for success and error cases.

regiskuckaertz:
I've done all the work but the XHR are not authorised for some reason, need to speak with the platform team

GHaberis:
@regiskuckaertz OK cool, I'll take a look at this PR now. Regarding the XHR post, are you seeing `nextgen` return a `403` forbidden error in response to your request?

regiskuckaertz:
Yeah, it fails during pre-flight.

GHaberis:
Hmm OK I've not attempted an XHR request directly from the templates before, however I wonder of this could be because the origin (the template) has the `file://` protocol.

regiskuckaertz:
I'm running the local dev on http://localhost:8000 😞

regiskuckaertz:
Perhaps because the target is https ... I'll try http instead.

regiskuckaertz:
No luck. I can only test if running on https myself. Perhaps the local dev environment could use a local certificate.

ArticleTemplates/assets/js/bootstraps/campaign.js
```diff
@@ -0,0 +1,68 @@
+define([], function () {
+    'use strict';
+
+    const rootPath = 'https://code.api.nextgen.guardianapps.co.uk/formstack-campaign/submit'
```
GHaberis: Unfortunately `const` only has partial support on older versions of Android that we're still supporting, so we'll need to convert to `var`. The same goes for `let` which has no support on older versions.

ArticleTemplates/assets/js/bootstraps/campaign.js
```diff
@@ -0,0 +1,68 @@
...
+      }
+
+      window.addEventListener('online', hideOfflineMessage.bind(null, campaign));
+      window.addEventListener('offline', displayOfflineMessage.bind(null, campaign));
+
+      campaign.addEventListener('submit', function (e) {
+        e.preventDefault();
+        const data = new FormData(campaign);
+        const json = {};
+        for(let [key, value] of data.entries()) {
```
GHaberis: I'd be wary about using destructuring assignments, I'm not sure older versions of android will support them.

ArticleTemplates/assets/js/bootstraps/campaign.js
GHaberis: The JS validation (`yarn validate`) will complain about a missing semicolon here as well.

ArticleTemplates/assets/scss/modules/_campaign.scss
```diff
@@ -0,0 +1,226 @@
...
+    border-top: 1px $garnett-neutral-4 solid;
+    list-style: none;
+
+    &::-webkit-details-marker {
+        display: none;
+    }
+
+    .campaign--kicker {
+        display: flex;
+        flex-direction: row;
```
GHaberis: Unfortunately older versions of Android (4.3 and below) require `-webkit-prefix` for `flex` and these needed to be added manually in these styled 😞 

`display: flex;`
`display: -webkit-flex;`
`flex-direction: row;`
`-webkit-flex-direction: row;`

regiskuckaertz: Ok I'll do that 😭

regiskuckaertz:
I've recently added something in the JSON payload from MAPI that this PR will need. What is the recommended way to access it?

ArticleTemplates/assets/scss/modules/_campaign.scss
regiskuckaertz: I've just used autoprefixer 😈

GHaberis:
@regiskuckaertz it'll require some native work. You could add a new placeholder in the `<script>` in which `GU.bootstrap.init` is called. The native layer could replace the placeholder with a stringified version of the JSON, which you can then parse. Something like:

HTML template

`var myJson = JSON.parse("__MY_JSON___")`

On app the native layer swaps the placeholder for the JSON string so this becomes:

`var myJson = JSON.parse("{ ... }")`;

alfavata:
@regiskuckaertz Can you access [our JIRA ticket](https://theguardian.atlassian.net/browse/GLA-500)? If so it'd be great if you could add there all the details about the placeholder and what we need to replace it with.

regiskuckaertz:
@GHaberis @alfavata thanks for your help 👌 

- I'll add the placeholder in the `init` object, let's say: `calloutsUrl: "__CALLOUTS_URL__"`
- sadly I don't have access to your JIRA so can't do anything 😐

alfavata:
@regiskuckaertz That's alright, I'll edit the ticket.

If I read the [MAPI changes](https://github.com/guardian/mobile-apps-api/pull/1154) correctly, the value to replace the placeholder with is in the `item.metadata.journalism.campaignsUrl` field of the JSON payload, right?

regiskuckaertz:
@alfavata yes, precisely 👍

regiskuckaertz:
... then let's be consistent and use `__CAMPAIGNS_URL__` to avoid confusion

regiskuckaertz:
Ok cool this is ready to go, whenever apps have been updated with the new field. cc @GHaberis @alfavata

GHaberis:
@regiskuckaertz OK great will give this a 👀again, I'm at a conference today, but will hopefully find time!

regiskuckaertz:
Thanks @GHaberis ! There's no rush, enjoy your conference :)

regiskuckaertz:
Used JSON in the end, not a big deal

PR #794, opened 2018-06-11
Bundle JS/CSS with Webpack

GHaberis:
The `.js`/`.css` assets are now bundled with Webpack. Babel transpiles the JS so we can now use modern JS language features.

The changes to the JS are the switch to using ES6 `import`/`export`. I wanted to keep touching everything in between to a minimum. 

I have however switched to named exports for tree shaking, so for example where we'd previously have imported `util` and called `util.signalDevice()`, we now `import { signalDevice } from 'util';` and call `signalDevice();`.

The AMD entry points `app.js` and `main.js` have been removed, and replaced with a new single entry point `boot.js`.

The other changes are to html templates to include these scripts in the head.

`<script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/curl.js"></script>`
`<script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/boot.js"></script>`

`boot.js` is the entry point for the app. `curl.js` is required for embedded interactives which are AMD modules.

Some template include the additional scripts:

`<script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/mobile-range-slider.js"></script>`
`<script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/smooth-scroll.js"></script>`

These are 3rd party libraries, that define `MobileRangeSlider` and `SmoothScroll` on the window.

TODO:
- Re-enable eslint (this will be done on a PR per file basis)
- Re-enable unit tests
- Re-enable local dev server with browsersync
- Re-enable .scss validation
- Remove eslint overrides in .eslintrc.js


GHaberis:
I'm closing this pull request but keeping the branch alive incase a @webb04 finds it a useful reference.

PR #795, opened 2018-06-13
Fix issues with embed video links

GHaberis:
Non-youtube video embeds in articles are broken on iOS and Android. 

This is because they need to be wrapped in an `<a>` tag. All `<a>` in the body currently have a style applied to them from the `text-underline` mixin, this applies a nice underline effect on links, but it shouldn't be used on `video-url` links, so I've put a `not(.video-url)`  exception on this mixin, which fixes the issue.

Before...

screenshot 6

After...

screen shot 2018-06-13 at 12 07 47

### DIFF
--- a/ArticleTemplates/assets/js/modules/youtube.js
+++ b/ArticleTemplates/assets/js/modules/youtube.js
@@ -254,8 +254,7 @@ define([
     }
 
     function onPlayerPlaying(id) {
-        var placeholderParent,
-            currentTime = Math.round(players[id].player.getCurrentTime());
+        var placeholderParent;
 
         stopPlayers(id);
         setProgressTracker(id);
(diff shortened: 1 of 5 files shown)


ArticleTemplates/assets/js/modules/youtube.js
```diff
@@ -254,8 +254,7 @@ define([
     }
 
     function onPlayerPlaying(id) {
-        var placeholderParent,
-            currentTime = Math.round(players[id].player.getCurrentTime());
+        var placeholderParent;
```
GHaberis: This is an unrelated javascript validation fix that was flagged when I ran `yarn validate` locally.

ArticleTemplates/assets/scss/garnett-modules/content/_meta.scss
```diff
@@ -103,7 +103,7 @@
         margin-left: base-px(.5);
         border-left: 1px solid color(brightness-86);
         margin-bottom: base-px(.5);
-        margin-top: base-px(-0.5);
+        margin-top: base-px(-.5);
```
GHaberis: This is an unrelated scss validation fix that was flagged when I ran `yarn validate` locally.

PR #798, opened 2018-06-26
AA-73 - Implement Line Height in Templates

ChrisOwen101:
This work is related to this ticket in the Android board:

https://theguardian.atlassian.net/browse/AA-73

This adds the ability to call `lineHeightResize()` from native iOS or Android code which allows the line spacing for the body text in articles to be changes at runtime. 

This is my first foray into the templates in any substantial way so let me know if there's any gotchas that I might have fallen in to, however from my testing everything seems to be working correctly.

Screenshots: https://photos.app.goo.gl/AHv7PfYZNqZp5dez8
1. Shows the new dialog with the slider
2. Shows the default line spacing (the same as previously)
3. Shows the maximum line spacing

### DIFF
--- a/ArticleTemplates/articleTemplate.html
+++ b/ArticleTemplates/articleTemplate.html
@@ -32,7 +32,6 @@
             sectionTone: "__SECTION_TONE__",
             isAdvertising: "__IS_ADVERTISING__" ? true : false,
             fontSize: "__FONT_SIZE__",
-            lineHeight: "__LINE_HEIGHT__",
             platform: "__PLATFORM__",
             isOffline: "__IS_OFFLINE__" ? true : false,
             contentType: "article",
--- a/ArticleTemplates/articleTemplate.html
+++ b/ArticleTemplates/articleTemplate.html
@@ -16,7 +16,7 @@
     <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/polyfills/requestAnimationFrame.js"></script>
     <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/bootstrap.js"></script>
 </head>
-<body class="garnett--type-__GARNETT_TYPE__ garnett--pillar-__GARNETT_PILLAR__ tone--__SECTION_TONE__ __FONT_SIZE__ __PLATFORM__ __IS_OFFLINE__ advert-config--__ADS_CONFIG__ __IS_ADVERTISING__ __DISPLAY_HINT__" data-content-type="article" data-ads-enabled="__ADS_ENABLED__" data-ads-enable-hiding="__ADS_ENABLE_HIDING__" data-ads-config="__ADS_CONFIG__" style="margin-top: __ACTIONBARHEIGHT__px;" data-font-size="__FONT_SIZE_INT__" data-template-directory="__TEMPLATES_DIRECTORY__" data-page-id="__PAGE_ID__" data-mpu-after-paragraphs="__MPU_AFTER_PARAGRAPHS__" data-use-ads-ready="__ADS_FAST_CALLBACK__" data-content-section="__SECTION__">
+<body class="garnett--type-__GARNETT_TYPE__ garnett--pillar-__GARNETT_PILLAR__ tone--__SECTION_TONE__ __FONT_SIZE__ __LINE_HEIGHT__ __PLATFORM__ __IS_OFFLINE__ advert-config--__ADS_CONFIG__ __IS_ADVERTISING__ __DISPLAY_HINT__" data-content-type="article" data-ads-enabled="__ADS_ENABLED__" data-ads-enable-hiding="__ADS_ENABLE_HIDING__" data-ads-config="__ADS_CONFIG__" style="margin-top: __ACTIONBARHEIGHT__px;" data-font-size="__FONT_SIZE_INT__" data-template-directory="__TEMPLATES_DIRECTORY__" data-page-id="__PAGE_ID__" data-mpu-after-paragraphs="__MPU_AFTER_PARAGRAPHS__" data-use-ads-ready="__ADS_FAST_CALLBACK__" data-content-section="__SECTION__">
 
     __ARTICLE_CONTAINER__
 
@@ -32,6 +32,7 @@
             sectionTone: "__SECTION_TONE__",
             isAdvertising: "__IS_ADVERTISING__" ? true : false,
             fontSize: "__FONT_SIZE__",
+            lineHeight: "__LINE_HEIGHT__",
             platform: "__PLATFORM__",
             isOffline: "__IS_OFFLINE__" ? true : false,
             contentType: "article",
--- a/ArticleTemplates/articleTemplateContainerFeatures.html
+++ b/ArticleTemplates/articleTemplateContainerFeatures.html
@@ -44,7 +44,7 @@
         __SPONSORSHIP__
 
         <div class="article__body" id="feature-body">
-            <div class="from-content-api prose resizable selectable" id="feature-body-blocks">
+            <div class="from-content-api prose resizable resizable_line_height selectable" id="feature-body-blocks">
                 __BODY__
             </div>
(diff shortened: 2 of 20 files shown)


mohammad-haque:
In my novice eyes all looks fine but it would be very good to review by @GHaberis

duarte:
Hey, can you add some more detail on this? I'm interested in understanding the design implications before we merge it

ChrisOwen101:
@duarte I've just added screenshots and some details of what it shows

ArticleTemplates/articleTemplateContainerImmersive.html
```diff
@@ -33,7 +33,7 @@ <h1 class="article__headline headline selectable">
     </div>
 
     <div class="article__body" id="article-body">
-        <div class="from-content-api prose resizable selectable">
+        <div class="from-content-api prose resizable resizable_line_height selectable">
```
maxspencer: Nit: check whether this should be with hyphens or not, most class names seem to use them, although `article__body` above doesn't...

ArticleTemplates/assets/scss/modules/_resizable.scss
```diff
@@ -19,3 +19,11 @@
         }
     }
 }
+
+.resizable_line_height {
+    @for $i from 0 through 10 {
+        .line-height-#{$i} & {
+            line-height: 1.5 + ($i * 0.1);
```
maxspencer: Does this preserve the existing default line height?

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -417,6 +423,15 @@ define([
         document.body.dataset.fontSize = replacementInt[2];
     }
 
+    function lineHeightResize(current, replacement) {
```
GHaberis: Can you give examples of `current` and `replacement` so I can see what class names are being swapped in/out?

ArticleTemplates/articleTemplate.html
```diff
@@ -16,7 +16,7 @@
...
+<body class="garnett--type-__GARNETT_TYPE__ garnett--pillar-__GARNETT_PILLAR__ tone--__SECTION_TONE__ __FONT_SIZE__ __LINE_HEIGHT__ __PLATFORM__ __IS_OFFLINE__ advert-config--__ADS_CONFIG__ __IS_ADVERTISING__ __DISPLAY_HINT__" data-content-type="article" data-ads-enabled="__ADS_ENABLED__" data-ads-enable-hiding="__ADS_ENABLE_HIDING__" data-ads-config="__ADS_CONFIG__" style="margin-top: __ACTIONBARHEIGHT__px;" data-font-size="__FONT_SIZE_INT__" data-template-directory="__TEMPLATES_DIRECTORY__" data-page-id="__PAGE_ID__" data-mpu-after-paragraphs="__MPU_AFTER_PARAGRAPHS__" data-use-ads-ready="__ADS_FAST_CALLBACK__" data-content-section="__SECTION__">
```
GHaberis: Can you give an example of what `__LINE_HEIGHT__` would be replaced with? I'm not sure if it's necessary to add it to the `<Body>` here.

ArticleTemplates/articleTemplateContainerFeatures.html
```diff
@@ -44,7 +44,7 @@ <h1 class="headline selectable" id="headline">__TITLE__ __REVIEW_RATING__</h1>
         __SPONSORSHIP__
 
         <div class="article__body" id="feature-body">
-            <div class="from-content-api prose resizable selectable" id="feature-body-blocks">
+            <div class="from-content-api prose resizable resizable_line_height selectable" id="feature-body-blocks">
```
GHaberis: Can you rename any references to the class `resizable_line_height` to `resizable-line-height`? (Single hyphens should be used as word separators in class names).

ArticleTemplates/assets/scss/modules/_resizable.scss
```diff
@@ -19,3 +19,11 @@
         }
     }
 }
+
+.resizable_line_height {
```
GHaberis: The css generated from this will  apply the appropriate line height to descendant elements of `.resize_line_height` which have the class `line-height-x`.

I can't see from the HTML templates when this would occur as we're never adding the class `line-height-x` to any of the descendant elements of `.resize_line_height`?

ArticleTemplates/assets/scss/modules/_resizable.scss
ChrisOwen101: As I understand it, this should preserve the default line height at 1.5.

ArticleTemplates/articleTemplateContainerImmersive.html
GHaberis: This should be with hyphens, BEM (http://getbem.com/introduction/) is the naming convention used in the CSS. `article__body` is actually a "block" `article` and "element" `body`, which are separated by double underscores.

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -417,6 +423,15 @@ define([
...
+        var replacementStr = replacement,
+            replacementInt = replacementStr.split('-');
+
+        document.body.classList.remove(current);
+        document.body.classList.add(replacement);  
+        document.body.dataset.lineHeight = replacementInt[2];
```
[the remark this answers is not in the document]
ChrisOwen101: We call `javascript:fontResize()` in the native layer whenever the user changes the font size in the dialog window.

ArticleTemplates/assets/js/bootstraps/common.js
ChrisOwen101: Current could be: `line-height-0`
Replacement could be: `line-height-1`

ArticleTemplates/articleTemplate.html
ChrisOwen101: `__LINE_HEIGHT__` could be replaced by `line-height-0`

duarte:
Thanks for adding more info Chris!

We've had a discussion about font settings a while back, and were hoping to move font sizing settings to system level. Unlike share and Save for Later, font settings are not something that needs to be set on a per-article basis. They're also not a control users change very often. In addition to this, system level font sizing is causing misalignment issues on articles on Android.

I'm also unsure about the need for the granularity of a slider for this. Do we have any more information on granularity from the accessibility survey? From a design perspective, the drawback of the granularity is that it destroys our hope of vertical rythm / baseline grid in the layout. If looser line spacing was implemented as an on-off switch, it would allow us to retain baseline grid consistency.

ArticleTemplates/articleTemplate.html
GHaberis: OK, I've seen with @ChrisOwen101 that the `.scss` compiles to `line-height-x .resize_line_height` so this is OK!

ChrisOwen101:
@duarte Very happy to iterate on this and I share your thoughts on moving the `Text Settings` button out of the article and in to settings. 

As for moving the font size to the system level (assuming by system level you means from the Android settings) is that this will change the font size in the entire app. This change _only_ changes the font size for the body fo the article.

For granularity, the survey doesn't say any more of the matter. I'd be happy to have fewer options which are more inline with our design guidelines if you let me know what they are. 

If you'd like to chat, come over to my desk and we can find the best solution for all of us. For now, are you alright for this feature to be added to the app with the idea that we'll move to a more exact solution in the future?

ArticleTemplates/collectionTemplate.html
```diff
@@ -28,6 +28,7 @@
         GU.bootstrap.init({             
             asyncStylesFilename: "__ASYNC_STYLES__",
             fontSize: "__FONT_SIZE__",
+            lineHeight: "__LINE_HEIGHT__",
```
GHaberis: You've removed `lineHeight: "__LINE_HEIGHT__"` from `articleTemplate.html`, I'd also do the same to the remaining instances in the HTML templates.

PR #802, opened 2018-07-09
Related cards position poller

GHaberis:
Related cards are native, however they rely on the templates accurately reporting the position of a placeholder in the template. This position is reported when the page initially loads, however the position of the placeholder can subsequently change as the DOM changes.

An implementation of `window.MutationObserver` was setup to handle this however it's been inconsistent and not always reported the updated position.

This PR removes that and replaces with a poller which reports the position of the placeholder to the native layer (provided it's changed), the interval gets progressively longer, up to a max interval of 4s.

The poller is reset when new liveblog blocks are loaded and new tweets are enhanced (as both these events could affect the position of the placeholder beneath it).


ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -424,9 +424,6 @@ define([
     }
 
     function lineHeightResize(current, replacement) {
-        var replacementStr = replacement,
```
GHaberis: Unrelated validation fix

ArticleTemplates/assets/scss/garnett-modules/_related-content.scss
```diff
@@ -7,4 +7,5 @@
         }
         width: 1200px;
     }
- }
\ No newline at end of file
+ }
+
```
GHaberis: Unrelated validation fix

ArticleTemplates/assets/scss/helpers/_variables.scss
```diff
@@ -30,4 +30,4 @@ $gs-gutter: 20px;
 $gs-span: 24.75rem;
 $gs-span-wide: 30.75rem;
 $fs-main-link: 100%;
-$fs-sub-link: 1.5rem;
\ No newline at end of file
+$fs-sub-link: 1.5rem;
```
GHaberis: Unrelated validation fix

ArticleTemplates/assets/scss/modules/_resizable.scss
```diff
@@ -23,7 +23,7 @@
 .resizable-line-height {
     @for $i from 0 through 10 {
         .line-height-#{$i} & {
-            line-height: 1.5 + ($i * 0.1);
+            line-height: 1.5 + ($i * .1);
```
GHaberis: Unrelated validation fix

PR #805, opened 2018-07-17
Define and expose function for allowing Live Blog block to be scrolled to smoothly.

aodhol:
This PR adds a function that will be called by the native layer passing the id of a Live Blog block in order to scroll that block into view. The [SmoothScroll](https://github.com/cferdinandi/smooth-scroll) library needed to be used here because [Safari on iOS doesn't yet support smooth scrolling](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) when using `window.scrollIntoView()`


alfavata:
Waiting on this PR to be merged before reviewing https://github.com/guardian/ios-live/pull/3989 -- thanks for your help @GHaberis :)

PR #807, opened 2018-07-23
Fix button styling when not in pillar and data wiring

regiskuckaertz:
screen shot 2018-07-23 at 16 54 28

- there was no styling by default when not in a pillar: it shouldn't happen, it does on the test mobile because they use a weird pillar called "test" or something like that 😅 
- added CORS headers in API gateway
- fixed the wiring of checkbox values


alfavata:
Sorry @regiskuckaertz, I could check the result with you but it's better if the code is looked at by someone that knows what they're doing 🙈

PR #809, opened 2018-07-27
Swap out const for var

GHaberis:
I noticed we had some es6 `const`s in the code, we don't transpile the templates so these need to be `var`.


regiskuckaertz:
Thanks @GHaberis

PR #810, opened 2018-07-27
add shouldRun check to cards position poller

GHaberis:
adds a `shouldRun` check to the `cards.js` module. `shouldRun` is `true` when an element with the class `.related-content` exists and the user is on iOS.  

We check if `shouldRun` is `true` before initialising the poller in `initPositionPoller`, previously we'd start the poller whether this condition was true or not. This wouldn't cause a bug, but it is needless code execution.

### DIFF
--- a/ArticleTemplates/assets/js/modules/cards.js
+++ b/ArticleTemplates/assets/js/modules/cards.js
@@ -6,22 +6,32 @@ function (
 ) {
     'use strict';
 
-    var initialised = false;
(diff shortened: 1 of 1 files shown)


GHaberis:
@webb04 I was going to close this but having reviewed the code it looks like it's still useful.

`initPositionPoller` is called on `iOS` and `Android` (on Tweet enhancements, new Liveblog blocks etc). This stops the poller running on `Android` as its not needed.

PR #812, opened 2018-07-30
Replace outdated font files

duarte:
Updates Guardian Headline font files to the latest version…

CC @paperboyo

PR #813, opened 2018-07-30
Adds italic font files for GH Guardian Headline

duarte:
While doing #812 I noticed it would be nice to be able to use italics in the app. This should only add 200kb (zipped size) to the bundle, but might add 500k if they're not zipped during the build.

@maxspencer will check for bundle size change compared to master on Android
@alfavata is happy with this for the iOS side


paperboyo:
Just one observation: we load very limited set of these on dotcom. As soon as the artists get their hands on more styles, they gonna use them, creating stuff that won’t be possible to replicate on all platforms. This is a strong argument in my mind to not load everything (apart from bundle size). @blongden73?

duarte:
@paperboyo this shouldn't be a problem. We have quite different designs on the app and the web, and we have different features. We don't necessarily intend on converging in all cases. Having extra font weights is a useful option for the apps.

blongden73:
@paperboyo I think this is ok...just because we still have control over which one gets used. I think there are specific app scenarios where we may want to use other weights. The reason we don't load them all on the website is more for size than anything else. If we could load more in we would.

blongden73:
👍 I'm cool with this.

PR #811, opened 2018-07-27
Enable CORS

regiskuckaertz:
I removed that line by mistake 😐 Sorry about that, @SLambrakis @mohammad-haque

PR #814, opened 2018-07-31
Revert "Enable CORS"

ajwl:
Reverts guardian/mobile-apps-article-templates#811

Testing from @SLambrakis and the Android team suggests that this change has stopped submit working on the forms. 

If we want to use `req.withCredentials = true;` we need to set the `Access-Control-Allow-Origin` header to point to the app url. 
```
      req.setRequestHeader('Access-Control-Allow-Origin', 'https://guardian.co.uk');
      req.setRequestHeader('Vary', 'Origin');
```
We should return to the use of `withCredentials` before rolling this out to editorial

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/campaign.js
+++ b/ArticleTemplates/assets/js/bootstraps/campaign.js
@@ -44,7 +44,6 @@ define([
       req.open('POST', endpoint);
       req.setRequestHeader('Content-Type', 'application/json');
       req.setRequestHeader('Accept', 'application/json');
-      req.withCredentials = true;
       req.onload = displayConfirmation.bind(null, campaign, form);
       req.onerror = displayError.bind(null, campaign, form);
       req.send(JSON.stringify(data));


ajwl:
@SLambrakis have made the revert - will just ask @GHaberis before merging

SLambrakis:
Thanks Anna! I'll test this now (cc @mohammad-haque )

SLambrakis:
Works for me, thank you - this should be sufficient for us to get a build out. Happy to do more testing in the future if/when you move back to `withCredentials`

mohammad-haque:
@SLambrakis  @ajwl 
I have just tested this on android and it works - but the time between clicking on `Submit` button and seeing the message `Thank you for your contribution` is between 10-15 seconds and without any kind of progress message/ui.

ajwl:
@mohammad-haque Hm - that's a good point. Also a problem on web.
Simplest solution would be to show some kind of waiting gif until the success response comes back. 
Better longer term strategy would be to rewrite the Lambda in javascript instead of scala - theres a long warm-up time with scala which delays things

mohammad-haque:
@ajwl Thanks. Good to know that scala takes longer to bootup.

PR #815, opened 2018-07-31
update sports sponsor background colour

webb04:


webb04:
Applied correct colours to sponsorship section.

Trello card: https://trello.com/c/V7ezjE64/305-background-colour-on-sponsor-logo-container-wrong-on-sport-liveblog-app-templates

Before:
image

After:
screen shot 2018-07-31 at 15 46 35

PR #825, opened 2018-08-13
keep contrast on dead blogs

webb04:
Still some tweaks required (increasing contrast more, especially the more button)

Before:
img_4637 2

After:
screen shot 2018-08-13 at 15 13 01


webb04:
Second commit for the same issue seen on live blogs

screen shot 2018-08-14 at 11 06 30

PR #833, opened 2018-08-29
avoid flash of alt comment on every article view

webb04:
- Initially set alt attribute to be colour of background to avoid flash of content

# TODO
- Set colour back to original once img is loaded
- Show article caption by default if offline?
- Stop "Match Info" temporarily showing


duarte:
Looks good from design perspective 👍

PR #835, opened 2018-08-30
avoid states where only half the space is used

webb04:


webb04:
There's more we could do on the apps. The web use some more colours and an image:

web_suggested

PR #836, opened 2018-08-30
increase width for large tablet screens

webb04:
iPad pro landscape screen width not currently handled:

# Before
screen shot 2018-08-30 at 15 07 55

# After
screen shot 2018-08-30 at 15 06 04


duarte:
Hey, we should try to get that `article-kicker` block (i.e 'Cricket / Fourth Test Match') left-aligned with the rest of the page, in the same place as 'Day 1 / The Ageas Bowl'

PR #843, opened 2018-09-06
Use GuardianTextEgyptian everywhere (AA-561 and GLA-633)

maxspencer:
This PR builds upon #842, it:

* includes the GuardianTextEgyptian font files in the templates repo,
* modifies the SCSS to use those fonts on both platforms,
* removes their old equivalents (`egyptiantext-*.otf`).

This PR should not be merged until iOS are happy with the changes and have had chance to test them.


webb04:
@maxspencer should this be merged?

maxspencer:
@webb04 I think so, but I'm confused because it seems that the previous PR #842 has been reverted now (looking at the diff here). It would still be right to check that this change is OK with iOS as it effects them.

JamieB-gu:
@maxspencer can we close this?

maxspencer:
@JamieB-gu what a blast from the past - Ideally we'd merge it to get rid of those old font names, but if it's causing you trouble then feel free to close.

JamieB-gu:
> It would still be right to check that this change is OK with iOS as it effects them.

@SLambrakis @faresite @dbk78 @aoifemcl15 what do you think?

maxspencer:
@JamieB-gu I am looking through my old open PRs in prep for leaving. I think we should merge this one, what do you think?

PR #854, opened 2018-10-03
Reposition ads and related content on layout changes

webb04:
toggle wouldn't work before iOS 10.3


webb04:
@duarte Do you know of any campaign examples? I think they are having a similar problem.

PR #859, opened 2018-10-08
iOS comment improvements

webb04:
| Before | After |
| --- | --- |
| screen shot 2018-10-08 at 14 13 57 | screen shot 2018-10-08 at 14 59 11 |
| screen shot 2018-10-08 at 14 14 21 | screen shot 2018-10-08 at 14 59 56 |

### DIFF
--- a/ArticleTemplates/assets/js/modules/comments.js
+++ b/ArticleTemplates/assets/js/modules/comments.js
@@ -206,10 +206,14 @@ define([
 
     function commentsClosed() {
         var commentsElem = document.getElementsByClassName('comments')[0],
-            discussionElem = document.getElementById('discussion');
+            discussionElem = document.getElementById('discussion'),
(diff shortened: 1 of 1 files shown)


alfavata:
Do you mean that this PR fixes https://theguardian.atlassian.net/browse/GLA-534? 😮

webb04:
I think it fixes one of those cases (when there's been no comments it shows "open for comments" but then it's closed).

The other one still needs fixing, the comment count is often different to the number we actually have. Need to have a deeper look at what's happening and make them both update at the same time.

alfavata:
I'm afraid the comment count issue has been there for a long time, but it's probably mostly due to caching. The ticket was specifically about the "Schrödinger's comments" issue so I closed it, we can create a new one if we still observe any misbehaviour.

PR #861, opened 2018-10-11
Remove hover states in articles

webb04:


webb04:
iPad pro content wasn't aligned, so some css updates included to fix
screen shot 2018-10-10 at 17 18 57

PR #871, opened 2018-10-24
iOS comment picks

webb04:


webb04:
screen shot 2018-10-30 at 18 18 43

ArticleTemplates/assets/js/modules/comments.js
```diff
@@ -266,6 +270,15 @@ define([
         }
     }
 
+    function showStaffBadges() {
```
GHaberis: So `showStaffBadges` sets `display: none` on all `.comment__role` elems where `getAttribute('data-role')` is falsy? Could you just do this with css instead by using a data-attribute selector? eg.
css
```
.comment__role[data-role=''] {
  display: none;
}
 ```

Then you could remove this JS.

webb04:
There's a case where responses can be selected as a Guardian pick. We should add badges to the picked comment (main comment or response) and we should scroll down to comment in the webview (populate id of each comment with commentId)

PR #872, opened 2018-11-08
Updated deprecated title piece font

mohammad-haque:
This PR updates the old title piece font with a new font. I have made change only for android and left ios unchanged for now.

This change request originally came from @paperboyo :

> Hi,
> 
> It was brought to my attention, some repos still contain old version of the Titlepiece font. Also, some contain GHF Display Figures font 😨. Both fonts are a danger to humanity and should be deleted from all repos.
> 
> Please find attached:
> – GT Guardian Titlepiece Bold in Full version (allowing you to switch default Oldstyle figure style to Lining in CSS to replace GHF Display Figures: same number design).
> 
> Thank you for your consideration!
> Thank you for your cooperation!
> 
> Regards


duarte:
Mohammad, could you please check where this is being used within the templates repo, so I can review design changes needed, if any? Should we review the impact on native in a separate PR, and time the merging of both so they come out in the same release?

mohammad-haque:
@duarte I'm working on android to make similar change and yes, they both needs to coordinate/sync because on android we use fonts from template/asset folder.

I've searched for old `titlepiece` font and I found only one reference and that is what I have changed in this PR. What I can see `front__header__title` css is using that font. 

@webb04 do you think there could be more usages?

duarte:
Thanks Mohammad — let's update that to use 'Guardian Headline' Bold, instead of titlepiece!

mohammad-haque:
@webb04 Anything on fronts is native and does not use html template. So I don't know what exactly you mean by football fixture on fronts.

I intend to keep this PR limited to font change only and any subsequent changes/refactoring we can do in a separate PR.

mohammad-haque:
@duarte this screen is proper native :)

PR #875, opened 2018-11-19
Related content and outbrain cleanup

webb04:
- Update background of related content + outbrain to match native iOS colour
- Update related content + outbrain borders to match fronts
- Use the same font and size for both titles
- Remove more hover states in outbrain container
- Fix alignment of guardian labs outbrain on iPad pro

| Before | After |
| --- | --- |
|  |  |
|  |  |

### DIFF
--- a/ArticleTemplates/assets/scss/base/_reset.scss
+++ b/ArticleTemplates/assets/scss/base/_reset.scss
@@ -18,7 +18,7 @@ audio, video {
 
 body,
 html {
-    background-color: color(brightness-93);
+    background-color: color(brightness-96);
     color: color(brightness-7);
     width: 100%;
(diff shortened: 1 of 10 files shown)


duarte:
Hey, I think this could be nicer if, on tablet, we make the outbrain container title go above the cards (like in related content), so the cards can align to the grid like they do on the related content container.

webb04:
Yeah I think that would look nicer (and also setting related content to the same font size). 
But we also have comments and comment picks in the same format, so I might make another ticket for that
screen shot 2018-11-20 at 10 42 59

webb04:
With the screenshot above, we might need to change the body colour too (uses of brightness-93 to be updated to brightness-96)

PR #878, opened 2018-11-20
Allow single taps on twitter embeds

webb04:
Seems like the 300ms tap delay was removed in iOS 9.3 (we support iOS 10+ now). https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away

For some reason, fastClick is causing a bug in liveblogs where you need to tap twice on twitter embeds for it to be recognised.

We could probably remove fastClick altogether now...


duarte:
This sounds great to me, but will leave code for @GHaberis to have a look at, as I don't completely understand it :)

PR #881, opened 2018-11-23
Add loading state once audio requested

webb04:
Using the existing loading state to show content is loading (this can be quite long, especially on slow 3g/4g connections).

The logic to add a loading state is there so we can update the design to anything else.

Could also animate the play/pause icons to make it less jumpy.

| Current | Suggested |
| --- | --- |
| current | after |

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/audio.js
+++ b/ArticleTemplates/assets/js/bootstraps/audio.js
@@ -78,6 +78,9 @@ define([
         }
 
         slider1.setValue(current);
+        if (current > 0) {
+            audioPlaying();
+        }
     }
 
     function audioSlider() {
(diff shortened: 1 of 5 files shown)


duarte:
Hey, I think we shouldn't use the roundel here. We should remove the icon, and add a scale animation to the yellow circle (maybe two circles or something). If we hold this, I'll try to do somethin for it next week!

webb04:
Yeah sure, the roundel did look a bit weird there but couldn't find any other loading states within app or website

duarte:
I have an idea for something new we can do with that circle that might be cool

duarte:
I've added the animation now :)
Can we test it on the simulatrion on your laptop?

For reference, it should look like this — https://uploads.guim.co.uk/2018/12/13/Screenflick_Movie_2.mp4

webb04:
It looks good! Me and Jorge spotted a bug with the earlier code changes, maybe we could fix that?

webb04:
We've also got to merge Jorge's PR in

webb04:
There's some js erroring, trying to find out where 😅

webb04:
Ah that's just going between webpack/grunt... looks good! but no longer pulses

PR #882, opened 2018-11-23
Update labs to new background

webb04:
| Before | After |
| --- | --- |
|  |  |
|  |  |


webb04:
@duarte do we need a darker colour at the top like in liveblogs?
screen shot 2018-11-23 at 10 26 13

duarte:
@webb04 no need for the header stuff!
But, can you please and update our variable for the dark background colour to be #4B8878? (it should replace the colour used for links, and for the background of the horizontal bar at the top). This will make it match the native layer changes in MSS-506

webb04:
Updated template variable, but looks like the header is native
screen shot 2018-11-23 at 11 04 27

PR #889, opened 2018-12-13
Correct embed widths

webb04:
| Before | After |
| --- | --- |
|||
|||
|||

### DIFF
--- a/ArticleTemplates/assets/scss/modules/external/_element.scss
+++ b/ArticleTemplates/assets/scss/modules/external/_element.scss
@@ -65,7 +65,8 @@
         overflow-x: hidden;
     }
 
-    .element-instagram .instagram-media {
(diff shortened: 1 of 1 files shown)


ArticleTemplates/assets/scss/modules/external/_element.scss
```diff
@@ -59,6 +59,17 @@
...
 }
+@include mq($to: col2) {
+    .element-embed,
+    .element-instagram {
+        overflow-x: hidden;
+    }
+
+    .element-instagram .instagram-media,
+    .element-embed iframe {
+        width: calc(100vw - 24px) !important;
```
duarte: Should this be base-px(2) instead, so we could potentially switch it to 8pt grid easier later?

PR #893, opened 2018-12-14
Comment icons

webb04:

### DIFF
--- a/ArticleTemplates/assets/scss/modules/_comments.scss
+++ b/ArticleTemplates/assets/scss/modules/_comments.scss
@@ -157,7 +157,7 @@
     &__count {
         font-family: $egyptian-display;
         color: color(brightness-60);
-
+        margin-left: 4px;
         .comments--page & {
             @include headline(.5);
             margin-bottom: base-px(1);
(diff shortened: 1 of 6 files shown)


duarte:
I have a coupel of things to check on this one:

* Is the tappable area of the buttons we've reduced to 32x32 still larger? If not can we use padding or similar to make the tappable area 48x48 centered around the circle
* There's too much space around the comment count. Can we move the 'upvote' icon to the far right (12px from the edge of the comment's box), and then have 12px space (and use base-px for these, we can change them to 8px later), and then the comment count number (right aligned)?
* For the 'view more' button, can we bring the text closer to the button by about 4px, and move the text 1px up (the fonts arent well aligned vertically, so they need a little help with padding:S)

webb04:
- Tappable area is 56 x 56
- Repositioned comment count - `flex-direction: column` and `order: 1` is cool!
- Updated upvote and view more colours

screen shot 2018-12-17 at 17 38 12
screen shot 2018-12-17 at 17 37 50

webb04:
Do you know the code for the right arrow icon or how to find if we have one? can remove the 90° rotation.

webb04:
Right icon
screen shot 2018-12-18 at 11 05 56

PR #930, opened 2019-01-29
Use sentry JavaScript error reporting

webb04:


webb04:
DSN does seem public for browsers:
https://github.com/guardian/support-frontend/blob/7324a13d6ea717906a3960c4beaff2c49a49fa63/assets/helpers/logger.js#L7

https://forum.sentry.io/t/raven-js-configure-lets-dsn-be-exposed-in-page-source/1816

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -45,6 +45,12 @@ function init() {
     initRichLinks();
     setupForms();
 
+    initSentry({
+        transport: Transports.FetchTransport,
+        dsn: 'https://8abc43d4e79b425eb6d4b5659ccd4020@sentry.io/40557',
+        release: `${BUILD_NUMBER}`
```
jorgeazevedo: Can you help me understand what the interpolation is doing here? I'm not sure how webpack works, but it looks like `BUILD_NUMBER` is a variable available in the scope. Is the interpolation being done as way to make `BUILD_NUMBER` a string?

webb04: I think ```release: BUILD_NUMBER``` should work the same way, but I saw they used a string in their [documentation](https://docs.sentry.io/workflow/releases/?platform=javascript#configure-sdk) so this interpolation is only converting it to a string.

jorgeazevedo: Ah thanks for clarifying, I see what you mean. In that context it might be worth changing the default value to `"0"` in the `DefinePlugin` call to make it a bit more explicit. Env variables are strings, so unless `process.env` is trying to be clever it should always return strings too.

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -45,6 +45,12 @@ function init() {
     initRichLinks();
     setupForms();
 
+    initSentry({
+        transport: Transports.FetchTransport,
+        dsn: 'https://8abc43d4e79b425eb6d4b5659ccd4020@sentry.io/40557',
```
SiAdcock: Which Sentry project does this refer to?

webb04: The [webview](https://sentry.io/the-guardian/webview/) project that was used for apps 3/4 years ago. Still getting a lot of reports from old installs 😅

SiAdcock: Interesting, thanks @webb04!

PR #931, opened 2019-01-30
Remove Yarn

webb04:
The Snyk team city job is failing because the two lock files are out of sync.

When installing a new module we have to install using yarn and npm.

NPM now has a `lock` file and is bundled with node so makes installing the repo slightly easier. 

screen shot 2019-01-30 at 12 39 17


webb04:
This explains it https://medium.com/@scinos/de-duplicating-yarn-lock-ae30be4aa41a

PR #936, opened 2019-02-05
Hide empty comments message if there's comments

webb04:


alfavata:
I've built this branch using the `templates-build-s3` [TeamCity job](url
https://teamcity.gutools.co.uk/viewType.html?buildTypeId=Apps_Templates_TemplatesS3v2) and tested it on a simulator, however I'm not sure how to find an article affected by the issue 🤔

PR #940, opened 2019-02-08
Add base rule for SVG

regiskuckaertz:
I would love to give a before/after preview but I've been struggling with making the dev environment work for a whole day and now I want to jump by the window.

So, the _only_ thing I was able to do, with numerous manoeuvres was to simulate the master version of the Android app. Then I navigated to the mystery bird article where the sample audio atom is. There I noticed a CSS rule:

```css
svg {
  position: absolute;
  right: 0;
}
```

That rules does not appear in the Web version so it must be coming from the app itself. Because of that rule, the layout of the audio player inside the atom was completely broken. After adding the fix below in the dev tools editor, the layout was correct.

Sadly, I have been completely unable to run the Android app by adding a local dependency on this project, the following error kept showing up:
```
> Could not list contents of '/Users/regis_kuckaertz/Projects/android-news-app/android-news-app/node_modules/@guardian/mobile-apps-article-templates/node_modules/eslint/node_modules/.bin/strip-json-comments'. Couldn't follow symbolic link.
```

I don't have time to debug. @webb04 btw can you please update the README, instead of manually updating the `package.json` file, one should run this instead:
```
npm i ../../mobile-apps-article-templates 
```

### DIFF
--- a/ArticleTemplates/assets/scss/base/_global.scss
+++ b/ArticleTemplates/assets/scss/base/_global.scss
@@ -8,8 +8,3 @@ body {
     font-size: 1.6em;
     line-height: 1.5em;
 }
-
-// svgs are positionned absolutely in android 🤔
-svg {
-    position: static;
-}
--- a/ArticleTemplates/assets/scss/base/_global.scss
+++ b/ArticleTemplates/assets/scss/base/_global.scss
@@ -8,3 +8,8 @@ body {
     font-size: 1.6em;
     line-height: 1.5em;
 }
+
+// svgs are positionned absolutely in android 🤔
+svg {
+    position: static;
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/core": {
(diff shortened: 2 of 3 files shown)


regiskuckaertz:
[In-Article Audio Atom ready to go live on one article (v2)](https://trello.com/c/sh46jpiC/598-in-article-audio-atom-ready-to-go-live-on-one-article-v2)

webb04:
Sorry that the process of setting up the dev environment isn't in the best state between the native repos and templates. I'll make sure to speak to the Android/iOS teams to see if we can make it easier for people to contribute.

We could look into bringing the test/fixtures back but I think there's a lot of benefits of testing on Simulators.

I've experienced the same svg issue on this article, it must be something leaking css on the page as I haven't seen it on any other article.   

I had the symbolic link issue before too. I think I solved the problem by trying one of the suggestions on this page https://github.com/facebook/react-native/issues/11212.

regiskuckaertz:
> I've experienced the same svg issue on this article, it must be something leaking css on the page as I haven't seen it on any other article.

Hmm. In any case this is clearly an app-specific behaviour: do you think this fix is necessary then? At least until we plug the leak.

ajwl:
Cant preview... but makes sense to me 👍

regiskuckaertz:
@webb04 I tried removing `node_modules` and the `.bin` folders. It does not work. I would be very much in favour of a development environment where we don't need to fight the build system, I've lost a whole day.

NataliaLKB:
Ah strange problem! Change seems sensible to me as a quick fix, and all you can really do without being able to test properly.

NataliaLKB:
@webb04 
> We could look into bringing the test/fixtures back but I think there's a lot of benefits of testing on Simulators.

True, you should definitely test on a simulator before publishing/merging to master, but testing locally should be a quick and easy experience. The test templates helped a lot with this. 

Totally your choice, but the risk of not making it easy to contribute is we rely on you a lot more to help us test.

regiskuckaertz:
I found where the nasty rule is coming from: the chart atom that is currently embedded inside the page. I've removed it from the test article, this fix is not necessary anymore.

PR #946, opened 2019-02-21
Pinch and zoom galleries

webb04:
| Before | After |
| --- | --- | 
|  |  |


ArticleTemplates/assets/js/bootstraps/gallery.js
```diff
@@ -1,120 +1,114 @@
...
+import Hammer from 'hammerjs';
 
 function init() {
-    buildGallery();
-    window.addEventListener('resize', debounce(buildGallery, 100));
-}
+    lazyLoadImages();
+    var galleryImages = document.querySelectorAll('.touch-gallery__images');
+    galleryImages.forEach(function(galleryImage) {
+        var hammerSettings = (GU.opts.platform === 'ios' ? { touchAction: 'auto' } : {})
```
GHaberis: `hammerSettings` will be the the same for each `galleryImage` so you can move this declaration outside the `forEach` and only run this once.

ArticleTemplates/assets/js/bootstraps/gallery.js
```diff
@@ -1,120 +1,114 @@
...
-    ];
+    });
 
-    // loop through images and add to rows
-    for (i = images.length - 1; i >= 0; i--) {
-        // if current row is full create new row
-        if (rows[0].length === galleryProps.imagesPerRow) {
-            rows.unshift([]);
+    function lockArticleSwipe(toggle) {
+        var isAndroidApp = (window.location.origin === "file://" && /(android)/i.test(navigator.userAgent) ) ? true : false;
```
GHaberis: Why are we testing if we're on the android app using this technique? Does `GU.opts.platform === 'android'` not behavew as you'd expect?

ArticleTemplates/assets/js/bootstraps/gallery.js
```diff
@@ -1,120 +1,114 @@
...
-    }
+      const handleIntersection = (entries) => {
+        entries.forEach(entry => {
+          if (entry.intersectionRatio > 0) {
+            entry.target.style.backgroundImage = `url(${entry.target.dataset.src})`;
+          }
+        })
+      }
+
+      const observer = new IntersectionObserver(handleIntersection, options);
```
GHaberis: I'm not sure about the minimum versions of iOS/Android that are supported but you may want to check whether `IntersectionObserver` works as expected on them.

ArticleTemplates/assets/scss/modules/_comments.scss
```diff
@@ -167,6 +167,10 @@
         font-family: $egyptian-display;
         color: color(brightness-60);
         margin-left: 4px;
+        @include mq($from: col4) {
+            margin: 4px 0px;
```
GHaberis: You don't need the `px` after `0`.

ArticleTemplates/assets/js/bootstraps/gallery.js
GHaberis: OK forget this ☝️, I've just seen you're using a polyfill.

ArticleTemplates/assets/js/bootstraps/gallery.js
webb04: Duarte had written that code before leaving so I'm not sure if there's a specific reason.
I've checked and `GU.opts.platform` returns "android" as expected so I've updated it.

PR #950, opened 2019-02-26
Add bootstrap for charts

regiskuckaertz:
This is a temporary hack required for launching chart atoms in the UI. They currently are added in the form of an iframe with local source:
```html
<iframe srcdoc="@atomHtml"></iframe>
```
and inside that code is some JavaScript that is sending messages to the parent document. This must be controlled in such a way that we have a listener able to handle those messages before they are sent. Hence the local complexity.

- [x] Tested with Android Studio
- [ ] Tested on iOS (cc @webb04 can you help with this?)

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/atoms.js
+++ b/ArticleTemplates/assets/js/bootstraps/atoms.js
@@ -11,6 +11,47 @@ function init() {
             return;
         }
         bootAtomType(t, atomTypes[t]);
+        if (t === 'chart') {
+            initCharts();
+        }
+    });
+}
+
+function initCharts() {
+    var iframes = Array.prototype.slice.call(document.querySelectorAll('.atom--chart > .atom__iframe'));
+    window.addEventListener('message', function (event) {
+        var message;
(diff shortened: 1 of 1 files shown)


regiskuckaertz:
[Charts in atom workshop (proof of concept of 1 type of chart)](https://trello.com/c/LotlAMdR/563-charts-in-atom-workshop-proof-of-concept-of-1-type-of-chart)

regiskuckaertz:
[Charts in apps](https://trello.com/c/Rjut1lzj/625-charts-in-apps)

regiskuckaertz:
screen shot 2019-03-01 at 13 25 22

Even though the chart looks borked, the fact that it's displayed and the frame has a correct height is a proof that this PR is correct. We can then get it merged (pending iOS testing); I will fix the invalid rendering of the chart itself in a separate project.

webb04:
Lgtm, feel free to merge if they look as expected!

iOS screenshots:
screen shot 2019-03-01 at 17 17 56
screen shot 2019-03-01 at 17 15 52
screen shot 2019-03-01 at 15 33 23

PR #979, opened 2019-04-01
Resizable text on media pages

webb04:
| Before | After |
| --- | --- |
|||


webb04:
Need to add a new `resizable-media` mixin based on 1.6rem font size and 2.0 line-height.

Add this class to Galleries, Videos and Audio pages.

PR #980, opened 2019-04-02
Use MutationObserver to notify native layers

webb04:
| Before | After |
| --- | --- |
|||

To test:
- [x] Tweets enhancing
- [x] Tags
- [ ] Videos on sports tabs
- [x] Atom explainers
- [ ] Number of times apps are being called
- [x] Quizes
- [ ] Interactives

Todo:
- [x] Simplify three notification functions (remove the timeouts)


ArticleTemplates/assets/js/modules/youtube.js
```diff
@@ -79,19 +76,6 @@ function checkForVideos() {
```
GHaberis: I wonder if this `resize` event listener is required now? Will the `MutationObserver` register a `resize` event? I believe `resize` was meant to pickup a rotation of device

ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -678,6 +683,19 @@ function setupForms() {
     }
 }
 
+function notifyNativeLayers() {
+    const targetNode = document.getElementsByTagName('body')[0];
```
GHaberis: Could be simplified to `const targetNode = document.body`

GHaberis:
@webb04 this is great, It simplifies things so much having a single `MutationObserver` compared to having to know when to call the old "poller" functions when ever the DOM's modified.  I haven't use `MutationObserver` before so don't have experience of what triggers the handler? When new liveblocks are added for example they're added to the DOM and then animated into position over 600ms. Will the `MutationObserver` handler run when they've settled into their permanent position?

PR #982, opened 2019-04-03
Only style byline links

webb04:
| Before | After |
| --- | --- |
|||

### DIFF
--- a/ArticleTemplates/assets/scss/pillar/_colour.scss
+++ b/ArticleTemplates/assets/scss/pillar/_colour.scss
@@ -294,7 +294,6 @@
         }
 
         // Author names
-        .byline,
         .headline__byline,
         .byline__author,
         .byline__author a,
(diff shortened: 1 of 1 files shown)


webb04:
@benwuersching I thought this could be a bug as it was quite hard to see what author had a link but it is intentional. 

It seems to be a bigger problem on media pages where we have a lot of authors.

Is there a better way to make the links clearer?

benwuersching:
@webb04 Hmmm, I checked web and app and noticed that we're styling byline links in a somewhat whimsical way (bold = link, italic = no link) and also inconsistently across different page types. On web when there's a black background, we do white byline links with underlining. On the app, when there's a black background, we have so far done the links in pillar colour without underlining. Everywhere else on web and app, we do the links in pillar colour without underlining. 

So we could do what you suggest for black background: grey text when something is not a link, otherwise links in pillar colour. I just find it a bit busy with the two colours. Alternatively, we could leave everything in pillar colour but underline the links.

author_links

PR #985, opened 2019-04-05
Remove Sentry

webb04:
We can put this in a conditional that's populated by the native layer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -950,63 +950,6 @@
         "@types/yargs": "^12.0.9"
       }
     },
-    "@sentry/browser": {
-      "version": "4.5.3",
-      "resolved": "https://registry.npmjs.org/@sentry/browser/-/browser-4.5.3.tgz",
-      "integrity": "sha512-ePbsyXtZS9jWM0lD8rxA46omJjJcQPaRZak08TyxFBGm1YOUFvvNAA6rNjUtK4fFEznLhPA5YcllcOCBE2bPPg==",
-      "requires": {
-        "@sentry/core": "4.5.3",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -11919,7 +11862,8 @@
     "tslib": {
       "version": "1.9.3",
       "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.9.3.tgz",
-      "integrity": "sha512-4krF8scpejhaOgqzBEcGM7yDIEfi0/8+8zDRZhNZZ2kjmHJ4hv3zCbQWxoJGz1iw5U0Jl0nma13xzHXcncMavQ=="
+      "integrity": "sha512-4krF8scpejhaOgqzBEcGM7yDIEfi0/8+8zDRZhNZZ2kjmHJ4hv3zCbQWxoJGz1iw5U0Jl0nma13xzHXcncMavQ==",
+      "dev": true
```
jorgeazevedo: hm is this dev property supposed to be here?

jorgeazevedo:
Yay! I think this will be very easy to revert and re-enable later on

package-lock.json
webb04: Should be ok, just a result of removing sentry I guess!
https://docs.npmjs.com/files/package-lock.json#dev

jorgeazevedo: cool!

SiAdcock:
Have you discussed with InfoSec whether tomhis is even necessary? I can put you in touch with some people who I’m sure would be more than happy to look into this, as they’ve done so with the website in the past

PR #992, opened 2019-04-18
Cricket styles typo

webb04:
| Before | After |
| --- |  --- |
| | |


webb04:
@benwuersching @JamieKMorrison This seems like an obvious typo but the result doesn't make much sense to me. Does anyone know if the tea and cake is correct here?

PR #994, opened 2019-04-24
Add gallery credits

webb04:
| Before | After |
| --- | --- |
|||


webb04:
TODO:
- font underline 10% darker
- remove padding above credit
- make credit same size as caption

PR #996, opened 2019-04-25
Replace agate sans with Guardian sans

webb04:
| Before | After |
| --- | --- |
|||
|||
|||
|||
|||
|||
|||


benwuersching:
@webb04 Looks all good

PR #1029, opened 2019-05-28
Native http api

webb04:


webb04:
@ajwl I can't remember the exact request the callouts needed to make, but do you think something like this would work?

webb04:
Article with callout: https://www.theguardian.com/business/2019/jun/19/lloyds-staff-happy-with-chief-executives-63m-pay-antonio-horta-osorio

| Required fields | Confirmation | Failure (no connection) | 
| --- | --- | --- |
|  |  |  |

PR #1062, opened 2019-07-03
Add One More Ad Slot to Articles

ChrisOwen101:
This PR adds one more add slot that follows these rules:

- First advert appears after the (n)-th paragraph (in the current config this is the 3rd)
- Second advert appears after the (n*3)-th paragraph (in the current config this is the 9th)

Pending sign off and work from the android and ios teams


ChrisOwen101:
Ah Tablets! Good catch. I'll take a look at fixing that now.

PR #1064, opened 2019-07-05
Pinch and Zoom cartoons

webb04:
This pr replaces the cartoon iframe with the individual comic strip images.

I tried launching the iOS/Android image viewers using `x-gu://openimage/` but they only work when the displayImages are set in mapi. This could be possible but would require mapi or iOS and Android work.

Currently you can pinch and zoom individual comic strips, and we may want to add panleft and panright swiping.

We also have access to the desktop image, which might be nicer to pinch one image rather than multiple separate images. 

Now we have them as images in the html there's a lot we can do (pinching, buttons, new image viewer), I guess we should discuss what would be the best user functionality. @alfavata @glockett @zeek01 

Can test on:
https://www.theguardian.com/football/series/david-squires-on
https://www.theguardian.com/profile/first-dog-on-the-moon

### DIFF
--- a/ArticleTemplates/assets/js/article.js
+++ b/ArticleTemplates/assets/js/article.js
@@ -1,5 +1,5 @@
 import { init as commonInit } from 'bootstraps/common';
-import { init as articleInit } from 'bootstraps/article';
+import { init as articleInit, cartoonView } from 'bootstraps/article';
 import { init as atomsInit } from 'bootstraps/atoms';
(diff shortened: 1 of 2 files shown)


webb04:
Actually maybe this could be a mapi html cleaner to convert these iframes into images? but we would need to add them to `displayImages` @alexduf

alexduf:
Sure why not!

PR #1097, opened 2019-09-16
new inject creative api

webb04:
Requires iOS and Android changes to be merged at the same time.

### DIFF
--- a/ArticleTemplates/assets/js/modules/creativeInjector.js
+++ b/ArticleTemplates/assets/js/modules/creativeInjector.js
@@ -51,7 +51,7 @@ function injectEpicCreative(creativeContainer) {
     }
 }
 
-function injectCreative(title, body, firstButton, secondButton) {
+function injectEpic(title, body, firstButton, secondButton) {
     if (isOnline() && !document.getElementById('creative-container')) {
         const creativeContainer = document.createElement('div');
         creativeContainer.id = 'creative-container';
@@ -62,8 +62,8 @@ function injectCreative(title, body, firstButton, secondButton) {
 }
 
 function init() {
-    window.injectCreative = injectCreative;
-    window.applyNativeFunctionCall('injectCreative');
+    window.injectEpic = injectEpic;
+    window.applyNativeFunctionCall('injectEpic');
 }
 
 export { init };
--- a/ArticleTemplates/assets/js/modules/creativeInjector.js
+++ b/ArticleTemplates/assets/js/modules/creativeInjector.js
@@ -7,10 +7,10 @@ import {
 
 let impressionSeen = false;
 
-function buttonHtml(buttonText) {
+function buttonHtml(buttonText, buttonName) {
     return `
         <div>
-            <a class="epic-button" href="x-gu://creative_tap/premium">
+            <a class="epic-button" href="x-gu://creative_tap/${buttonName}">
                 ${buttonText}
                 <svg class="epic-button-arrow" xmlns="http://www.w3.org/2000/svg" width="23" height="23" viewBox="5 5 20 20">
                     <path fill="#121212" d="M22.8 14.6L15.2 7l-.7.7 5.5 6.6H6v1.5h14l-5.5 6.6.7.7 7.6-7.6v-.9"/>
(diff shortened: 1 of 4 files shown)


alexanderedge:
I see your comment on the JIRA ticket:

> We might need to check how impressions are tracked because we no longer have an id sent to the webView.

Would it be possible to pass the ID into `injectCreative(...)`?

webb04:
yep we can have another `ID` parameter and use it how we do currently e.g. `signalDevice("creative_impression/ID")`

webb04:
@alexanderedge I remember @SLambrakis saying we only have one Epic per page so iOS/Android would know what that ID was?

Happy to add it back if it's better to handle native side though.

alexanderedge:
> @alexanderedge I remember @SLambrakis saying we only have one Epic per page so iOS/Android would know what that ID was?
> 
> Happy to add it back if it's better to handle native side though.

Ah, okay. I was thinking of minimising changes iOS-side. I'll take a look tomorrow and see if we can get around not sending the ID into the webview.

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -67,41 +51,13 @@ function injectEpicCreative(creativeContainer) {
...
-
-    creativeContainer.id = id;
-    creativeContainer.classList.add('creative-container');
-    creativeContainer.classList.add(`${type}-creative-container`);
-    creativeContainer.innerHTML = html;
-
-    if (type === 'inline-article') {
-        injectInlineCreative(creativeContainer, 10);
-    } else {
+function injectCreative(title, body, firstButton, secondButton) {
```
alexanderedge: I was speaking with @SLambrakis and she mentioned that we only inject epics. Therefore, should this function be called `injectEpic(...)` to make it clear what it does?

webb04: yep can do, will be clear it's a new api too.

alexanderedge:
Unsure what the button title length requirements are but the arrow isn't vertically centred if it wraps:

Simulator Screen Shot - iPhone Xʀ - 2019-09-18 at 11 32 05

alexanderedge:
> @alexanderedge I remember @SLambrakis saying we only have one Epic per page so iOS/Android would know what that ID was?
> 
> Happy to add it back if it's better to handle native side though.

No need to add the ID back as there will only be a single creative.

webb04:
I'll remove the max-width on that button and change `injectCreative` to `injectEpic`

webb04:
Screen Shot 2019-09-18 at 12 35 26

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -5,58 +5,42 @@ import {
...
-    if (!trackedImpressions.includes(id) && isElementPartiallyInViewport(creativeContainer)) {
-        signalDevice(messageName);
-        trackedImpressions.push(id);
-    }
+let impressionSeen = false;
+
+function buttonHtml(buttonText) {
+    return `
+        <div>
+            <a class="epic-button" href="x-gu://creative_tap/premium">
```
JamieB-gu: Will this URL will be hit for both buttons on Android?

ArticleTemplates/assets/js/modules/creativeInjector.js
```diff
@@ -5,58 +5,42 @@ import {
...
-    let i;
-    let liveBlogEpicContainer;
-
-    for (i = 0; i < liveBlogEpicContainers.length; i++) {
-        liveBlogEpicContainer = liveBlogEpicContainers[i];
-        liveBlogEpicContainerId = liveBlogEpicContainer.getAttribute('id');
-
-        liveBlogEpicContainer.setAttribute('data-tracked', 'true');
+function isCreativeInView(creativeContainer) {
+    const messageName = `creative_impression`;
```
JamieB-gu: Could this be a normal string?

ArticleTemplates/assets/scss/modules/_injectedCreative.scss
```diff
@@ -1,19 +1,46 @@
...
+        color: color(brightness-7);
+        background: color(tone-highlight);
+        border-radius: 100px;
+        margin: 0 15px 15px 0;
+        padding: 6px 6px 4px 15px;
+        font-size: 16px;
+        font-weight: 900;
+        font-family: "Guardian Text Sans Web", "Helvetica Neue", Helvetica, Arial, "Lucida Grande", sans-serif;
+        line-height: 26px;
+    }
```
JamieB-gu: As discussed offline, font re-ordering.

PR #1100, opened 2019-09-30
liveblog.js: let native layers pass block id as-is

alfavata:
The newly-introduced (and not yet used on iOS) function `scrollToBlock` accepts a block id, but it's requiring the native layer to remove the `block-` prefix from the id... because the function itself prepends it to the supplied id!

That is, as far as I know, the only place where we're doing such a thing, hence this PR :)

[Agreed with @jordanterry that we can change the behaviour of the Android app to match]


alfavata:
Thanks - we just need for the changes in the Android code before we can merge this. I'll either wait for Jordan to be back or see if I can find the time to do them myself.

PR #1102, opened 2019-10-25
Hide all loading icons when new blocks are inserted

webb04:
@alfavata This has been pushed to https://s3-eu-west-1.amazonaws.com/builds.gutools.co.uk/guardian-mobile-apps-article-templates-v1.0.225.tgz.

Hopefully it works, if not there's two other cases we might need to hide the loading icons.


alfavata:
Unfortunately I can still see the loading indicator when it should be hidden:

IMG_EF01C17D73AA-1

alfavata:
This doesn't happen consistently: sometimes they're all gone, sometimes I can see 3 of them...

alfavata:
And this is the end state when there should be no more gap at all:

IMG_FED0BD29FCF6-1

alfavata:
Would it help if the native code called a JS function explicitly before inserting the new blocks?

alfavata:
(I think this change has made things worse, if anything - I wasn't seeing that many previosuly...)

alfavata:
@webb04 As discussed, please check out https://github.com/guardian/ios-live/tree/gla-1082-linking-to-liveblog-blocks if you want to experiment with this

alfavata:
The key event are using the colour only for live blogs, not dead ones - not sure if that's what @benwuersching wants?

benwuersching:
@webb04 @alfavata Whenever we show a list of key events linking to the referring blocks, then those links should be in colour (red for news pillar, blue for sport, gold for culture etc). This should be the case in 'live' blogs as well as 'dead' blogs.

alfavata:
I've tested this on Android too and it seems to work very well, I think it's ready to merge!

webb04:
Thanks for reviewing! I'll merge this on Monday and also add the link colours on dark mode.

PR #1103, opened 2019-10-30
Jw light immersive backgrounds

webb04:
This interactive had no background colour set so we were seeing the webView background colour.

Maybe it's best to keep the webView background light for immersive interactives?

| Before | After |
| --- | --- |
|||


ArticleTemplates/interactiveTemplate.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US">
+<html lang="en-US" style="background-color: #f6f6f6">
```
maxspencer: Should this be `#f6f6f6` rather than `#ffffff`?

PR #1109, opened 2019-11-15
Updated iOS Simulator Instructions - npm install

JamieB-gu:
## Why?

As far as I can tell, you need to run `npm install` to pick up the new (local) location of the templates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ $ npm install
     "@guardian/mobile-apps-article-templates": "file:../mobile-apps-article-templates"
 }
 ```
+* Run `npm install` in the root of the `ios-live` repo.
 
 ### Running on Android simulator
 * Checkout the branch you are developing against


alfavata:
I seem to remember that Xcode used to run `npm install` as part of the build process, then that was changed for some reason. Given @webb04 is away, maybe @jorgeazevedo and @jacobwinch will remember?

jorgeazevedo:
It still does I believe! `build-post` is still a thing right? https://github.com/guardian/ios-live/blob/e69db58b7ad23892e6e4cd515cfb0054483ea295/GLA/Scripts/build-post.sh#L8

You should be able to see the output of the `npm` call inside xcode, in the build log output section referring to the `build-post` script

JamieB-gu:
Ah ok, I wonder why that didn't happen when I first ran it... I got a bit muddled somewhere. Ok, thanks both, I'll close this.

PR #1122, opened 2019-12-09
Update webpack, and the terser and copy plugins

JamieB-gu:
## Why are you doing this?

There was [a vulnerability](https://github.com/advisories/GHSA-h9rv-jmmf-4pgx) reported in `serialize-javascript`, which appears in this project as a transitive dependency of webpack (via `terser-webpack-plugin`), `terser-webpack-plugin` and `copy-webpack-plugin`. This was fixed in patch releases for [v1](https://github.com/webpack-contrib/terser-webpack-plugin/releases/tag/v1.4.2) and [v2](https://github.com/webpack-contrib/terser-webpack-plugin/releases/tag/v2.2.2) of the terser plugin. I think the `copy-webpack-plugin` should pick up the patched release of `serialize-javascript` as of [v5.0.5](https://github.com/webpack-contrib/copy-webpack-plugin/compare/v5.0.4...v5.0.5).

**Note** I've updated across a major release (v1 -> v2) for our direct dependency on `terser-webpack-plugin`. This isn't strictly necessary because, as mentioned above, there *is* a patch in v1.4.2. I thought it was best to do this for future-proofing, as the breaking changes don't look [too dramatic](https://github.com/webpack-contrib/terser-webpack-plugin/releases/tag/v2.0.0). However, if anyone thinks it would be better to be more conservative and stick to v1 please comment.

For more info see the similar PR on [apps-rendering](https://github.com/guardian/apps-rendering/pull/103).

## Changes

- Bump `webpack`, `terser-webpack-plugin` and `copy-webpack-plugin` to their respective latest versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1883,175 +1883,179 @@
       "dev": true
     },
     "@webassemblyjs/ast": {
-      "version": "1.7.11",
-      "resolved": "https://registry.npmjs.org/@webassemblyjs/ast/-/ast-1.7.11.tgz",
-      "integrity": "sha512-ZEzy4vjvTzScC+SH8RBssQUawpaInUdMTYwYYLh54/s8TuT0gBLuyUnppKsVyZEi876VmmStKsUs28UxPgdvrA==",
+      "version": "1.8.5",
+      "resolved": "https://registry.npmjs.org/@webassemblyjs/ast/-/ast-1.8.5.tgz",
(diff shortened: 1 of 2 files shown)


JamieB-gu:
I'm seeing some weirdness running this with `npm run dev`. On app build in Xcode webpack does an incremental compile and throws an error. However, I can't see any impact in the simulator, and this *doesn't* happen for `npm run build`.

If anyone spots any issues please let me know.

PR #1126, opened 2020-02-07
audio atom and newsletter signup dark styles

webb04:
| Before | After |
| --- | --- |
|||
|||


webb04:
There's 44 other newsletter we would have to add css rules for... I can't see another way of doing this as there's nothing to differentiate these signups to other `element-embed`s

PR #1131, opened 2020-03-04
Fix Contrast On Atoms

JamieB-gu:
## Why?

@rupertbates reported low contrast on interactive atoms in dark mode. This should fix that, although I'm not sure if there's likely to be a knock-on effect of this change elsewhere.

## Changes

- Removed white colour from heading
- Applied dark colour to links

## Screenshots

| Before | After |
| --- | --- |
| dark-before | dark-fix |


JamieB-gu:
> Do we not need the `.heading` style?

Not sure... It didn't seem to cause any immediate problems, but that's not to say it won't have an effect somewhere else...

benwuersching:
@JamieB-gu Sorry I didn't get to look at this earlier. I'm happy with it, thank you.

PR #1136, opened 2020-03-16
media dark mode links

webb04:
- Increased line-height for text on gallery pages
- Use grey with underline for media links (dark mode)
- Remove underline in rich links (dark mode)

| Before | After |
| --- | --- |
|||
|||
|||


benwuersching:
@webb04 All good. Just one thing about the text links. Their colour should be the same as text colour but they're darker. Is this because they turned into visited links and are therefore dimmed?

benwuersching:
@webb04 Looking good now. Thanks very much.

PR #1137, opened 2020-03-18
lighter kicker colour

webb04:
| Before | After |
| --- | --- |
|||


benwuersching:
@webb04 This is perfect, thank you.

PR #1138, opened 2020-03-19
variable ad slots

webb04:
This pr sends an array of ad slot positions instead of comma separated string values for iOS and individual integers to Android.

Before we were showing two ads per article after the 3rd and 9th paragraphs, now we will show ads after the 3rd, 9th and after every following 6 paragraphs.
- This extends our existing logic but we can tweak it
- I've set a remote limit of 15 but this can be extended too
- `__MAXIMUM_ARTICLE_ADVERTS__` is the remote config value that needs passing from iOS and Android (I've raised a PR in mobile-static)

### iOS

I've not changed the names yet, but I think the function iOS calls `getMpuPosCommaSeparated` should be renamed to something like `getAdSlots`. We might also need to change to postMessage to be able to send arrays and objects from the webview to iOS: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

### DIFF
--- a/ArticleTemplates/articleTemplate.html
+++ b/ArticleTemplates/articleTemplate.html
@@ -54,7 +54,8 @@
             disableEnhancedTweets: "__DISABLE_ENHANCED_TWEETS__" === "true" ? true : false,
             atoms: atoms,
             hasEpic: "__HAS_EPIC__" === "true",
-            campaignsUrl: "__CAMPAIGNS_URL__"
+            campaignsUrl: "__CAMPAIGNS_URL__",
+            maximumAdverts: "__MAXIMUM_ARTICLE_ADVERTS__"
         });
     }());
     </script>
--- a/ArticleTemplates/assets/js/modules/ads.js
+++ b/ArticleTemplates/assets/js/modules/ads.js
@@ -140,16 +140,6 @@ function getMpuPosCommaSeparated() {
     return getMpuPos();
 }
 
-function getMpuOffset() {
-    return getMpuPos(getMpuOffsetCallback);
-}
-
-function getMpuOffsetCallback(params) {
-    return (numberOfMpus > 1)
-        ? `${params.x1}-${params.y1}:${params.x2}-${params.y2}`
-        : `${params.x1}-${params.y1}`;
-}
-
 function updateAndroidPosition() {
     if (adsType === 'liveblog') {
         getMpuPos(updateAndroidPositionLiveblogCallback);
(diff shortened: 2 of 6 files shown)


alexduf:
What if instead of having the template worrying about the maximum amount of ads the array was truncated in the native part of the app?

webb04:
@maxspencer I just had a test too and can see the growing misplacement as you scroll down. I can inspect it a bit more on Monday but I haven't changed the logic there yet, just the shape of it now being an array of objects.

This code can print out the templates current positioning (For https://www.theguardian.com/uk-news/2020/mar/12/millennium-dome-experience-disaster-inside-story-new-labour):

```
    const advertSlots = document.getElementsByClassName('advert-slot__wrapper');
    const scrollLeft = document.scrollingElement ? document.scrollingElement.scrollLeft : document.body.scrollLeft;
    const scrollTop = document.scrollingElement ? document.scrollingElement.scrollTop : document.body.scrollTop;
    const params = [];
    let advertPosition;

    if (advertSlots.length) {
        for (let i = 0; i < advertSlots.length; i++) {
            advertPosition = advertSlots[i].getBoundingClientRect();

            if (advertPosition.width !== 0 && advertPosition.height !== 0) {
                params.push({
                    x: advertPosition.left + scrollLeft,
                    y: advertPosition.top + scrollTop,
                    width: advertPosition.width,
                    height: advertPosition.height
                })
            }
        }

        console.log(params);
    }

0: {x: 0, y: 2082.5, width: 412, height: 250}
1: {x: 0, y: 4117.5, width: 412, height: 250}
2: {x: 0, y: 6730.5, width: 412, height: 250}
3: {x: 0, y: 9066.5, width: 412, height: 250}
4: {x: 0, y: 11125.5, width: 412, height: 250}
5: {x: 0, y: 13629.5, width: 412, height: 250}
6: {x: 0, y: 15592.5, width: 412, height: 250}
7: {x: 0, y: 18413.5, width: 412, height: 250}
8: {x: 0, y: 20052.5, width: 412, height: 250}
9: {x: 0, y: 22345.5, width: 412, height: 250}
```

webb04:
@maxspencer Aoife has seen something similar on iOS, so I'm looking into that now.

mohammad-haque:
@webb04 my local template build fails when I run `npm run build`. Is this change also available in S3 (i.g. https://s3-eu-west-1.amazonaws.com/builds.gutools.co.uk/.....)?

mohammad-haque:
> @webb04 my local template build fails when I run `npm run build`. Is this change also available in S3 (i.g. https://s3-eu-west-1.amazonaws.com/builds.gutools.co.uk/.....) so I can test using S3 instead of local build?

@webb04 Sorry. I forgot that I can build it in teamcity and grab s3 url :)

PR #1144, opened 2020-03-27
GLA-1880: fix Special Report article colours in dark mode

jeanlauliac:
I hate using `!important`. It's one of the most annoying feature of CSS. But I'm not sure I have too much choice here, because I use it only to override others, existing `!important`, located in `_darkModeArticle.scss`. (That's precisely what makes `!important` annoying: once we start using it, it proliferates.)

I don't want to touch/change *these*, in turn, because I suspect this would break other stuff in other places (otherwise they wouldn't be here in the first place), and I don't think it'd be worth to refactor the whole article CSSes just to get Special Report colours done.

So, here you go. These fixes are based on designs available from: https://theguardian.atlassian.net/browse/GLA-1880

You can repro locally by using CODE MAPI and opening the World/Europe front (PROD MAPI would work too I think).

@benwuersching would you like to have a look and give final approval?

| Before | After |
| --- | --- |
| Simulator Screen Shot - iPhone 11 - 2020-03-27 at 11 51 53 | Simulator Screen Shot - iPhone 11 - 2020-03-27 at 15 00 31 |

More screenshots of the new colours:

| Podcasts | Opinion | Immersive |
| --- | --- | --- |
|  Simulator Screen Shot - iPhone 11 - 2020-03-27 at 15 35 41 | Simulator Screen Shot - iPhone 11 - 2020-03-27 at 15 00 12 |  Simulator Screen Shot - iPhone 11 - 2020-03-27 at 15 00 00 |

### DIFF
--- a/ArticleTemplates/assets/scss/themes/darkMode/_darkModeSpecialReport.scss
+++ b/ArticleTemplates/assets/scss/themes/darkMode/_darkModeSpecialReport.scss
@@ -71,6 +71,16 @@ $specialReportBackground: rgb(34, 37, 39);
             background-color: transparent;
         }
 
+        .audio-player__button.touchpoint--primary .touchpoint__button,
+        .element-youtube .element-placeholder__button .touchpoint__button,
+        .audio-player__slider__knob {
+            background-color: $ratingYellow;
+
+            &:active {
+                background-color: $ratingYellow;
+            }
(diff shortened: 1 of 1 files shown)


benwuersching:
@jeanlauliac All looking good. Thank you.

PR #1145, opened 2020-03-30
glab dark mode fixes

webb04:
| Before | After |
| --- | --- |
|||
|||
|||
|||
|||


benwuersching:
@webb04 Thanks, that's good. I'm in the process of creating a few more tickets around G Labs (colour contrast issues etc.). But we'll deal with those later, separately.

PR #1146, opened 2020-03-30
center gallery adverts on tablets

webb04:
A recent change to ad slots caused this issue on gallery pages.

| Before | After |
| --- | --- |
|||


benwuersching:
Yes @webb04, that's much better 👍

PR #1158, opened 2020-04-22
remove styles from immersive photo essay articles

webb04:
These styles will still be on immersive articles, but not immersive photoEssay articles.

| Before | After |
| --- | --- |
|||


webb04:
Closing this as it probably makes sense to just release in apps-rendering.

There would always be some legacy users who would see some weird styles.

It's not broken now, but can be a nice enhancement as part of apps-rendering. It's actually already implemented.

PR #1160, opened 2020-04-27
Use the correct fonts in injected creatives

duarte:
Fix for the epic using the pre-garnett _Guardian Egyptian_ instead of _Guardian Headline_ for headlines, and _Guardian Headline_ instead of _Guardian Text Egyptian_ for body text.

<h3>Before</h3>

<h3>After</h3>

### DIFF
--- a/ArticleTemplates/assets/scss/modules/_injectedCreative.scss
+++ b/ArticleTemplates/assets/scss/modules/_injectedCreative.scss
@@ -13,7 +13,6 @@
         font-size: 20px;
         font-weight: 900;
         font-family: $egyptian-display;
-        color: tomato !important;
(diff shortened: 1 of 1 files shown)


webb04:
Spotted this today, but probably a little too new https://twitter.com/argyleink/status/1254794309263491072?s=20

duarte:
Thanks Jamie, great catch! I unnecessarily moved a margin around, so I can just put that back easily

duarte:
@webb04 I think it should be good now — can you check if it's ok with multiple buttons?

duarte:
Thanks Jamie!

SLambrakis:
cc @tombstonehand FYI

PR #1163, opened 2020-05-01
Compatibility with dotcom on 'Guardian Egyptian Web' font name

duarte:
Warning • This is confusing:

The app currently loads a font, 'Guardian Egyptian Web', which is no longer in use — and can be replaced by 'Guardian Headline' in all instances.

Unfortunately, dotcom loads 'Guardian Headline' under the 'Guardian Egyptian Web' name (i.e the wrong name, which is likely done for backwards compatibility). This leads to things on the app using the wrong font occasionally — mostly in things that are originally implemented for the web. We can fix them individually (like #1160) but we'd always be playing catch-up.

This PR makes 'Guardian Egyptian Web' an alias for 'Guardian Headline' for compatibility with dotcom. It doesn't delete the font files, in case they're needed by the native layer (another PR will follow for that).

### DIFF
--- a/ArticleTemplates/assets/scss/base/_typography.scss
+++ b/ArticleTemplates/assets/scss/base/_typography.scss
@@ -145,35 +145,29 @@ $fontMap: (
     // LEGACY FONTS:
 
     // Display Egyptian
-    GuardianEgyptianWeb100: (
-        family: 'Guardian Egyptian Web',
-        weight: 100,
-        android-src: 'url("../fonts/egyptiandisplay-thin.otf")',
-        ios-src: 'url("../fonts/egyptiandisplay-thin.otf")'
-    ),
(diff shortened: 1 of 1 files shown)


ArticleTemplates/assets/scss/base/_typography.scss
```diff
@@ -145,35 +145,29 @@ $fontMap: (
     // LEGACY FONTS:
 
     // Display Egyptian
-    GuardianEgyptianWeb100: (
```
webb04, line 148: Is there no interactives using this?

duarte: Yes there might be, but there's no _Thin_ weight for _Guardian Headline_, so it will automatically pick up _Light_ at weight 200

PR #1168, opened 2020-05-06
use white text for quiz message in dark mode

webb04:
| Before | After |
| --- | --- |
|||


benwuersching:
@webb04 Perfect, thank you!

PR #1169, opened 2020-05-12
Dropcap weight update

webb04:
- Removes border above first dropcap
- Increases font weight for dropcap
- Update border colour

| Before | After |
| --- | --- |
|||
|||
|||
|||


benwuersching:
Yes @webb04 , this looks very good to me. Thanks for picking it up so quickly.

PR #1170, opened 2020-05-18
hide advert variants

webb04:
Can be set using by updating `__TEST_SPEC__` on `articleTemplate.html`

```
"tests": "{
   hideAdsTest: 0
}"
```

| id | test |
| --- | --- |
| 0 | normal/control |
| 1 | different text |
| 2 | yellow banner |

| Light | Dark |
| --- | --- |
|||
|||
|||
|||

### DIFF
--- a/ArticleTemplates/assets/js/boot.js
+++ b/ArticleTemplates/assets/js/boot.js
@@ -97,7 +97,7 @@ const init = opts => {
         };
 
         const getHideAdsTest = (tests) => {
-            if (!tests) return 2;
+            if (!tests) return 0;
             const parsedTests = JSON.parse(tests);
             if (!parsedTests.hideAdsTest) return 0;
             return parsedTests.hideAdsTest;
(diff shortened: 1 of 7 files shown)


ArticleTemplates/assets/js/boot.js
```diff
@@ -96,19 +96,28 @@ const init = opts => {
             return 'default';
         };
 
+        const getHideAdsTest = (tests) => {
+            if (!tests) return 0;
+            const parsedTests = JSON.parse(tests);
```
JamieB-gu, line 101: Occurs to me that `JSON.parse` can throw an exception, is it worth wrapping in `try...catch`?

SLambrakis:
Very nit-picky, but there's a slight red line inside the button on the yellow banner (underneath the text). Probably not a thing any user would notice though!

benwuersching:
@SLambrakis Aha! Very well spotted, thank you, I missed that tiny detail. Of course, once you've seen the red line you can never unsee it. Would it be possible to get rid of it?

webb04:
Good spot, I've removed that red line now.

PR #1172, opened 2020-05-27
Narrower standfirsts

duarte:
This reapplies #1165 but now plays well with liveblogs and comment articles.

Comment — keylines still full-width

Liveblog — the narrow standfirst no longer applies to these:

(Edit fixed a :not, :not logic mistake)

### DIFF
--- a/ArticleTemplates/assets/scss/modules/content/_standfirst.scss
+++ b/ArticleTemplates/assets/scss/modules/content/_standfirst.scss
@@ -19,11 +19,15 @@
     font-weight: 400;
     @include standfirst();
     clear: left;
-    width: calc(100% - #{$gs-unit*6});
-    max-width: 600px;
 
     strong,
     b {
         font-weight: 700;
     }
+
+    body:not(.garnett--type-live) &,
+    body:not(.garnett--type-analysis) & {
+        padding-right: $gs-unit*6;
+    }
+
 }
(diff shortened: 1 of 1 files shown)


ArticleTemplates/assets/scss/modules/content/_standfirst.scss
```diff
@@ -24,4 +24,9 @@
     b {
         font-weight: 700;
     }
+
+    body:not(.garnett--type-analysis):not(.garnett--type-live) & {
```
webb04, line 28: Is it possible to have a comment article that isn't `analysis`?

webb04: Does `body` work? I thought that was targeting the `body` element which is outside of `.standfirst`?

duarte: What I'm trying to do is to exclude analysis (the articles with the underlined headlines like [this](https://www.theguardian.com/politics/2020/may/26/inconsistencies-between-cummings-lockdown-story-and-his-wifes)) and liveblogs.

Targetting `body` t works with the `&` after — so that applies to `body:not(.garnett--type-analysis):not(.garnett--type-live) .standfirst`!

duarte:
Good point, I'll check thru in dark mode!

duarte:
@webb04 I reviewed in dark mode and generally it seems okay to me! Is there anything specific within dark mode we should check?

duarte:
hey @webb04 does this look good to merge?

PR #1175, opened 2020-06-04
Dummy README Change

JamieB-gu:
## Why?

Testing a hypothesis about why builds are failing.


JamieB-gu:
Experiment successful. The test job now runs `npm run build && npm test`, instead of just `npm test`.

PR #1176, opened 2020-06-08
GLA-2087: add "Try Live" card

jeanlauliac:
Can we tested via instructions on https://github.com/guardian/ios-live/pull/5201

| On phone | On tablet |
| --- | --- |
| ||
|   |  |


ArticleTemplates/assets/js/bootstraps/liveblog.js
```diff
@@ -130,6 +130,22 @@ function onLiveMoreClick(liveMoreElem) {
...
+    for (let i = 0; i < elems.length; ++i) {
+        let elem = elems[i];
+        elem.addEventListener('touchstart', () => {
+            elem.classList.add('pressed');
+        });
+        elem.addEventListener('touchend', () => {
+            elem.classList.remove('pressed');
+        });
+        elem.addEventListener('click', () => {
+            signalDevice('try-live');
```
webb04, line 144: I think this might only be for iOS so we would need a platform check and probably a new JS interface in Android? @jordanterry

ArticleTemplates/assets/scss/modules/_live-promo.scss
```diff
@@ -0,0 +1,69 @@
+.block--live-promo {
+    background-color: color(tone-highlight);
```
webb04, line 2: Do we want the darker yellow in dark mode?

ArticleTemplates/assets/js/bootstraps/liveblog.js
jordanterry: I'll investigate this as soon as I get around to working on this

jeanlauliac: `signalDevice` is cross-platform, Android just need to respond to the same event. This is already used for the "show more" button so you can take this as example.

ArticleTemplates/assets/scss/modules/_live-promo.scss
jeanlauliac: Yes you're right! I'll add that

ArticleTemplates/assets/js/bootstraps/liveblog.js
jeanlauliac: So essentially I think we can merge this safely, since the card wouldn't be shown on Android for now anyway, and when it's integrated, Jordan can make sure that's handled as expected.

PR #1177, opened 2020-06-09
GLA-2087: Try Live: add close button

jeanlauliac:
When the button is pressed, we remove the card from the DOM. We also need to send a signal to the app so that we register the closure and we don't reopen the card. You can test this PR by following the steps on https://github.com/guardian/ios-live/pull/5203

| Before | After |
| --- | --- |
|||

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/liveblog.js
+++ b/ArticleTemplates/assets/js/bootstraps/liveblog.js
@@ -131,33 +131,31 @@ function onLiveMoreClick(liveMoreElem) {
 }
 
 function setupTryLive() {
-    const elems = document.getElementsByClassName('block--live-promo');
-    for (let i = 0; i < elems.length; ++i) {
-        let elem = elems[i];
+    const elem = document.getElementsByClassName('block--live-promo')[0];
+    if (elem == null) return;
 
-        let tryLiveButton = elem.getElementsByClassName('live-promo__button')[0];
-        tryLiveButton.addEventListener('touchstart', () => {
-            tryLiveButton.classList.add('pressed');
-        });
-        tryLiveButton.addEventListener('touchend', () => {
-            tryLiveButton.classList.remove('pressed');
-        });
-        tryLiveButton.addEventListener('click', () => {
-            signalDevice('try-live');
-        });
+    let tryLiveButton = elem.getElementsByClassName('live-promo__button')[0];
+    tryLiveButton.addEventListener('touchstart', () => {
+        tryLiveButton.classList.add('pressed');
+    });
+    tryLiveButton.addEventListener('touchend', () => {
+        tryLiveButton.classList.remove('pressed');
+    });
+    tryLiveButton.addEventListener('click', () => {
+        signalDevice('try-live');
+    });
 
-        let closeButton = elem.getElementsByClassName('live-promo__close-button')[0];
-        closeButton.addEventListener('touchstart', () => {
-            closeButton.classList.add('pressed');
-        });
-        closeButton.addEventListener('touchend', () => {
-            closeButton.classList.remove('pressed');
-        });
-        closeButton.addEventListener('click', () => {
-            elem.remove();
-            signalDevice('close-try-live');
-        });
-    }
+    let closeButton = elem.getElementsByClassName('live-promo__close-button')[0];
+    closeButton.addEventListener('touchstart', () => {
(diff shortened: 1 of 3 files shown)


ArticleTemplates/assets/scss/modules/_live-promo.scss
```diff
@@ -67,4 +68,31 @@
...
+        align-items: center;
+        justify-content: center;
+        padding: 0;
+
+        & div {
+            width: 14px;
+            height: 14px;
+
+            & svg {
+                display: block;
```
jeanlauliac, line 90: `svg` are supposed to be "block" by default, but this actually make the layout work, otherwise it's slightly misaligned. So… not sure. Might be a peculiarity of iOS browser. We can adjust this if we need to when implementing the Android version.

ArticleTemplates/assets/scss/modules/_live-promo.scss
```diff
@@ -67,4 +68,31 @@
...
+        border-radius: 1000px;
+        border: 1px solid #12121288;
+        position: absolute;
+        top: 12px;
+        right: 0;
+        width: 32px;
+        height: 32px;
+        display: flex;
+        align-items: center;
+        justify-content: center;
```
jeanlauliac, line 82: Leveraging flexbox to align the icon both vertically and horizontally.

ArticleTemplates/liveblogTemplateLivePromo.html
```diff
@@ -118,5 +118,12 @@ <h2 class="live-promo__title">Only got a minute?</h2>
                     fill="#333333" />
             </svg>
         </div>
+        <button class="live-promo__close-button">
```
webb04, line 121: I think we use a cross in an icon font elsewhere. The icons can have blank margins though so this is probably easier to style.

ArticleTemplates/assets/js/bootstraps/liveblog.js
```diff
@@ -131,18 +131,32 @@ function onLiveMoreClick(liveMoreElem) {
 }
 
 function setupTryLive() {
-    const elems = document.getElementsByClassName('live-promo__button');
+    const elems = document.getElementsByClassName('block--live-promo');
```
webb04, line 134: Is this remote configurable on how many "try live" blocks we show?

jeanlauliac: There should only ever be one on any liveblog, really. I'm happy to change this line to just pick the item zero.

webb04: could do `const elem = document.querySelector('.block--live-promo');` ?

ArticleTemplates/liveblogTemplateLivePromo.html
jeanlauliac: I wasn't able to find it. Happy to replace with a `data-icon` specifier if we find what is the code…

jeanlauliac: nevermind, found it!

ArticleTemplates/assets/js/bootstraps/liveblog.js
```diff
@@ -131,19 +131,31 @@ function onLiveMoreClick(liveMoreElem) {
...
-        });
-        elem.addEventListener('touchend', () => {
-            elem.classList.remove('pressed');
-        });
-        elem.addEventListener('click', () => {
-            signalDevice('try-live');
-        });
-    }
+    const elem = document.getElementsByClassName('block--live-promo')[0];
+    if (elem == null) return;
```
webb04, line 135:
```suggestion
    if (elem === undefined) return;
```

jeanlauliac: Sure thing. I general the JS community has settled on double-equal + `null` because it's a catch all for all "nullish" states, and it can be fiddly to make the difference. We may want to consider applying an automatic lint rule based on the direction we'd like (though of course, this codebase's going away with SSR ^_^ )

jeanlauliac: Ah actually `querySelector` returns `null`, so I'll go back to `null` ;)

jeanlauliac: Well at least in automated tests

PR #1178, opened 2020-06-10
GLA-2087: dismiss Try Live when pressed

jeanlauliac:
I forgot to add logic to dismiss the card when pressing the "Try Live" button itself. We do want to do this because once people went ahead and tried it we don't need to advertise it anymore, and it wouldn't show on subsequent LiveBlog renders. @jordanterry does it make sense for Android's logic as well?


webb04:
This may cause adverts and native videos to be out of position temporarily. But the poller should eventually reposition them.

We use these functions elsewhere to immediately reposition native elements https://github.com/guardian/mobile-apps-article-templates/blob/fcd8208a3e757bf85b5d16e96debe49ae75725e2/ArticleTemplates/assets/js/bootstraps/liveblog.js#L52-L54

PR #1198, opened 2020-09-02
Support iframe click tracking

alexduf:
This work is part of a larger piece to ensure the newsletter team closes the data gap we have as much as possible.

Part of that work is understanding when people are signing up to newsletters, which are currently implemented using iFrame embeds. Some work [has been done](https://github.com/guardian/frontend/pull/22806/files) on dotcom such that a message is posted from within the iFrame to the outer frame.

This PR implements a handler to receive the message, and forward it to the native layer. Unfortunately, the iOS stack blocks the message because of the security policy of the webview, so we'll only have Android support until the apps-rendering project reaches maturity.

See [this PR](https://github.com/guardian/android-news-app/pull/6241) for the android work that forwards the message to ophan

### DIFF
--- a/ArticleTemplates/assets/js/modules/ophan-iframe.js
+++ b/ArticleTemplates/assets/js/modules/ophan-iframe.js
@@ -12,7 +12,7 @@ const handleIframeMessage = (messageEvent) => {
 };
 
 const init = () => {
-    window.top.addEventListener('message', handleIframeMessage, false);
+    window.addEventListener('message', handleIframeMessage, false);
 };
 
 export {
--- a/ArticleTemplates/assets/js/modules/ophan-iframe.js
+++ b/ArticleTemplates/assets/js/modules/ophan-iframe.js
@@ -1,0 +1,20 @@
+const handleIframeMessage = (messageEvent) => {
+    const correctType = messageEvent.data.type === 'ophan-iframe-click-event';
+    const correctOrigin = messageEvent.origin === 'https://www.theguardian.com' ||
+        messageEvent.origin === 'https://m.code.dev-theguardian.com/';
+
+    if (correctOrigin && correctType) {
+        const clickEvent = messageEvent.data.value;
+        if (window.GuardianJSInterface && window.GuardianJSInterface.trackInPageClick) {
+            window.GuardianJSInterface.trackInPageClick(JSON.stringify(clickEvent));
+        }
+    }
+};
+
+const init = () => {
+    window.top.addEventListener('message', handleIframeMessage, false);
+};
+
+export {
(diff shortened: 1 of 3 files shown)


ArticleTemplates/assets/js/modules/ophan-iframe.js
```diff
@@ -0,0 +1,20 @@
...
+    if (correctOrigin && correctType) {
+        const clickEvent = messageEvent.data.value;
+        if (window.GuardianJSInterface && window.GuardianJSInterface.trackInPageClick) {
+            window.GuardianJSInterface.trackInPageClick(JSON.stringify(clickEvent));
+        }
+    }
+};
+
+const init = () => {
+    window.top.addEventListener('message', handleIframeMessage, false);
```
JamieB-gu, line 15: Do you need `.top` here? Isn't this code already running in the topmost frame?

ArticleTemplates/assets/js/modules/ophan-iframe.js
```diff
@@ -0,0 +1,20 @@
+const handleIframeMessage = (messageEvent) => {
+    const correctType = messageEvent.data.type === 'ophan-iframe-click-event';
+    const correctOrigin = messageEvent.origin === 'https://www.theguardian.com' ||
+        messageEvent.origin === 'https://m.code.dev-theguardian.com/';
```
JamieB-gu, line 4: Does this mean all `iframe` sources will be on the dotcom domain? As in, all relevant `iframe`s on the page will look something like:

```html
<iframe src="https://www.theguardian.com/..." />
```

JamieB-gu (commented):
Looks good! Couple of questions.

ArticleTemplates/assets/js/modules/ophan-iframe.js
alexduf: Yes, I want to make sure we don't accept any message from any origin (thinking about twitter embed and such)

ArticleTemplates/assets/js/modules/ophan-iframe.js
alexduf: I guess not? I'll change that :)

ArticleTemplates/assets/js/modules/ophan-iframe.js
JamieB-gu: So all newsletter embeds on apps are served from the dotcom domain?

JamieB-gu (approved):

alexduf:
Thanks Jamie! I've re-tested just to be sure and it's fine 👍

PR #1267, opened 2020-11-02
Premium taster ad slot

webb04:
### Description
---
Requires iOS branch: LIVE-616-iOS-taster-ad-slot

Inserts graphic into article in place of the first advert. When scrolled into view, the action is reported to iOS so that we don't show it for following articles.

Only supported on iPhone at the moment.

### TODO
---
- Do we need an Android equivalent to the iOS branch? @maxspencer @jordanterry 

### Screenshot
---
Screen Shot 2020-11-02 at 11 07 22

### DIFF
--- a/ArticleTemplates/assets/js/modules/premiumTaster.js
+++ b/ArticleTemplates/assets/js/modules/premiumTaster.js
@@ -1,0 +1,34 @@
+function init() {
+    try {
+        const showAdFreeTaster = window.GU.opts.showAdFreeTaster;
+        setInterval(() => {
+            console.log("showAdFreeTaster-------------------")
+            console.log(showAdFreeTaster);
+        }, 1000)
+        if (!showAdFreeTaster) {
+            return;
+        }
+
+        const afterParagraphs = 3;
+        const placeholder = document.createElement('div');
+        const adFreeTasterSibling = document.querySelector(`.article__body > div.prose > p:nth-of-type(${afterParagraphs}) ~ p + p`);
+
+        if (!(adFreeTasterSibling && adFreeTasterSibling.parentNode)) {
+            // Not enough paragraphs on page to add adFree taster
+            return;
+        }
+
+        setInterval(() => {
+            console.log("adFreeTasterSibling-------------------")
+            console.log(adFreeTasterSibling);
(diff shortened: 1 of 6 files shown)


ArticleTemplates/assets/js/modules/premiumTaster.js
```diff
@@ -0,0 +1,72 @@
...
+    const afterParagraphs = 3;
+    const placeholder = document.createElement('div');
+    const adFreeTasterSibling = document.querySelector(`.article__body > div.prose > p:nth-of-type(${afterParagraphs - 1}) ~ p + p`);
+
+    if (!(adFreeTasterSibling && adFreeTasterSibling.parentNode)) {
+        // Not enough paragraphs on page to add adFree taster
+        return;
+    }
+
+    const node = document.createElement('div');
```
JamieB-gu, line 40: Worth using a `DocumentFragment` for this?

ArticleTemplates/assets/scss/modules/_premium-taster.scss
```diff
@@ -0,0 +1,67 @@
...
+    margin: 12px 0;
+    display: inline-block;
+    border-radius: 50%;
+    background: color(brand400);
+    color: white;
+    line-height: 1;
+    box-sizing: content-box;
+
+    width: calc(100vw - 24px);
+    height: calc(100vw - 24px);
```
JamieB-gu, line 11: How does this look on tablet?

JamieB-gu (approved):

ArticleTemplates/assets/scss/modules/_premium-taster.scss
faresite: We are not running the test on tablets. phones only.

ArticleTemplates/assets/js/modules/premiumTaster.js
faresite: Makes sense 👍

faresite: @JamieB-gu TIL you can't set innerHtml onto a DocumentFragment.

PR #1284, opened 2020-11-16
opinion feature article dark mode styles

webb04:
This is a rare article combination that was released last week. Add the classnames to support dark mode styles.

https://www.theguardian.com/commentisfree/2020/nov/07/i-know-how-ill-get-through-this-winter-oversized-jumpers-and-triple-jacket-potatoes

| Before | After |
| --- | --- |
|||


JamieB-gu (approved):
A non-comment piece with the Opinion pillar - these things do exist!

PR #1285, opened 2020-11-16
Fix a11y for premium taster indicators

faresite:
Paired with @GiuliaAriu on this

The web premium taster indicators' accessibility was broken. This PR fixes it.

### DIFF
--- a/ArticleTemplates/assets/js/modules/premiumTaster.js
+++ b/ArticleTemplates/assets/js/modules/premiumTaster.js
@@ -71,8 +71,6 @@ function indicatorHtml(taster) {
         svg = offlineSvg
     }
 
-    window.a11yText = accessibilityText
-
     return `
         <span class="icon">${svg}</span>
         <section role="text" aria-label="${accessibilityText}">${text}</section>
--- a/ArticleTemplates/assets/js/modules/premiumTaster.js
+++ b/ArticleTemplates/assets/js/modules/premiumTaster.js
@@ -57,20 +57,25 @@ function init() {
 function indicatorHtml(taster) {
     if (taster !== "adFree" && taster !== "offline") return "";
     let text = ""
+    let accessibilityText = ""
     let svg = ""
     if (taster === "adFree") {
         text = "Experiencing the app ad free is a Premium feature. As a new user you can enjoy it for <strong>free for one week.</strong>"
+        accessibilityText = "Experiencing the app ad-free is a Premium feature. As a new user you can enjoy it for free for one week."
         svg = adFreeSvg
     }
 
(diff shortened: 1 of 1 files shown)


ArticleTemplates/assets/js/modules/premiumTaster.js
```diff
@@ -57,20 +57,25 @@ function init() {
...
     let svg = ""
     if (taster === "adFree") {
         text = "Experiencing the app ad free is a Premium feature. As a new user you can enjoy it for <strong>free for one week.</strong>"
+        accessibilityText = "Experiencing the app ad-free is a Premium feature. As a new user you can enjoy it for free for one week."
         svg = adFreeSvg
     }
 
     if (taster === "offline") {
         text = "Enhanced offline reading is a Premium feature. As a new user you can enjoy it for <strong>free for one week.</strong>"
+        accessibilityText = "Enhanced off-line reeding is a Premium feature. As a new user you can enjoy it for free for one week."
```
faresite, line 70: `reeding` isn't a typo here.. otherwise VoiceOver reads it like Reading, England .. 😏

faresite: Same with `off-line` and `ad-free` .. otherwise the intonation is broken too.

ArticleTemplates/assets/js/modules/premiumTaster.js
```diff
@@ -57,20 +57,25 @@ function init() {
...
         svg = offlineSvg
     }
 
+    window.a11yText = accessibilityText
```
webb04, line 74: What is this for?

faresite: 🤦‍♂️

ArticleTemplates/assets/js/modules/premiumTaster.js
webb04: That's really odd, surely reading is used loads more than Reading?
I wonder how Android pronounces it + would screenreader users understand _Reading_ if it's used elsewhere?

webb04 (approved):

ArticleTemplates/assets/js/modules/premiumTaster.js
JamieB-gu: That's bizarre. Is that VoiceOver? Sounds like a bug worth reporting?

faresite: It is odd indeed.. I can't find much about if that's a typo or not .. but things like Live we write as Lyve for a11y in iOS. Who would you report the bug with though?

PR #1360, opened 2021-04-06
[LIVE-2137] Remove smooth scrolling animations

maxspencer:
Users have written to us about the smooth scrolling animation we use (particularly for skipping to certain live blog blocks) being unsuitable.

Priscilla has confirmed we ought to remove this effect. This came through as an Android ticket, but I found it was actually a templates thing and thought I could probably fix it myself. I did a quick Google search and came up with using the built-in [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) function.

This PR comprises two commits:

1. removes the smooth scroll animation from live blogs as the ticket requires
2. removes one other remaining use of the `smooth-scroll` dependency and that dependency itself.

Please check I've done (2) correctly, but I have built and run this branch locally and it all seems OK!

Jira ticket: https://theguardian.atlassian.net/browse/LIVE-2137

| Before | After |
| --- | --- |
|before|after|

### DIFF
--- a/ArticleTemplates/assets/js/modules/quiz.js
+++ b/ArticleTemplates/assets/js/modules/quiz.js
@@ -465,8 +465,10 @@ function showScore() {
     initPositionPoller();
 
     // Scroll score panel into view
-    const scoresElement = document.getElementsByClassName('quiz-scores')[0]
-    scoresElement.scrollIntoView()
+    const scoresElement = document.getElementById('quiz-scores')
+    if (scoresElement !== undefined) {
+        scoresElement.scrollIntoView()
+    }
 }
 
 function showResult() {
--- a/ArticleTemplates/assets/js/modules/quiz.js
+++ b/ArticleTemplates/assets/js/modules/quiz.js
@@ -1,4 +1,3 @@
-import SmoothScroll from 'smooth-scroll';
 import { updateMPUPosition } from 'modules/ads';
 import { getStringFromUnicodeVal, getElementOffset } from 'modules/util';
 import { initPositionPoller } from 'modules/cards';
@@ -466,8 +465,8 @@ function showScore() {
     initPositionPoller();
 
     // Scroll score panel into view
-    const scroll = new SmoothScroll();
-    scroll.animateScroll('#quiz-scores', null, {speed: 1500, offset: 40});
+    const scoresElement = document.getElementsByClassName('quiz-scores')[0]
+    scoresElement.scrollIntoView()
 }
 
 function showResult() {
--- a/ArticleTemplates/assets/js/modules/util.js
+++ b/ArticleTemplates/assets/js/modules/util.js
@@ -1,5 +1,3 @@
-import SmoothScroll from 'smooth-scroll';
-
 function isElementPartiallyInViewport(el) {
     const rect = el.getBoundingClientRect();
     const windowHeight = (window.innerHeight || document.documentElement.clientHeight);
@@ -134,8 +132,7 @@ function getAndroidVersion() {
 }
 
 function scrollToElement(element) {
-    const scroll = new SmoothScroll();
-    scroll.animateScroll(element, { speed: 1500 });
+    element.scrollIntoView()
 }
 
 function isAdvertising() {
--- a/ArticleTemplates/simpleBodyTemplatePrepopHelpAndroid.html
+++ b/ArticleTemplates/simpleBodyTemplatePrepopHelpAndroid.html
@@ -135,13 +135,5 @@
 
         </div>
     </div>
-
-    <script type="text/javascript" src="__TEMPLATES_DIRECTORY__assets/build/smooth-scroll.min.js"></script>
-    <script type="text/javascript">
-        if (SmoothScroll) {
-            var smoothScroll = new SmoothScroll();
-            smoothScroll.init();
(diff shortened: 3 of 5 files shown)


prisalcalde:
@maxspencer thanks for attaching the videos. They are very helpful. I would suggest taking users directly to the Live blog entry without showing the top of it. Showing the top for just a few seconds may cause confusion. They don't have control over their interaction because it automatically takes them somewhere else.

frankie297:
Looks like a good idea to me, my only question or query is in the MDN docs this is classed as experimental, do we know what the means in terms of compatibility? @JamieB-gu Do you know the context of experimental in MDN's case? I couldn't find a good explanation. https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

maxspencer:
> @maxspencer thanks for attaching the videos. They are very helpful. I would suggest taking users directly to the Live blog entry without showing the top of it. Showing the top for just a few seconds may cause confusion. They don't have control over their interaction because it automatically takes them somewhere else.

@prisalcalde yes that would be good if we can do it. I think _some_ delay may be unavoidable if the block is not loaded in the initial 10 block range - but it's something we should investigate. Perhaps we could at least show a spinner or something so the user knows something is happening?!

However, that's not something we could tackle in this PR which only deals with the way the scrolling happens once it is triggered.

ArticleTemplates/assets/js/modules/quiz.js
```diff
@@ -466,8 +465,8 @@ function showScore() {
     initPositionPoller();
 
     // Scroll score panel into view
-    const scroll = new SmoothScroll();
-    scroll.animateScroll('#quiz-scores', null, {speed: 1500, offset: 40});
+    const scoresElement = document.getElementsByClassName('quiz-scores')[0]
```
JamieB-gu, line 468: The syntax previously used to select this suggests it's an `id` (the `#` tends to signify that), but you've switched to selecting it via class name. Are both present on the element?

ArticleTemplates/assets/js/modules/quiz.js
```diff
@@ -466,8 +465,8 @@ function showScore() {
...
+    scoresElement.scrollIntoView()
```
JamieB-gu, line 469: It's possible that `scoresElement` will be `undefined` if an element with that class (or `id` if you use `getElementById`) can't be found. If you call a method on `undefined` you'll get a runtime exception (because JavaScript), so you may want to wrap this call:

```suggestion
    if (scoresElement !== undefined) {
      scoresElement.scrollIntoView()
    }
```

JamieB-gu (commented):
Thanks for getting rid of this dep!

Just a couple comments.

> @JamieB-gu Do you know the context of experimental in MDN's case?

Discussed offline, but for anyone who's interested: we've concluded the use here is fine because the browser support is there. The experimental status may be due to the status of the spec, or could be related to the limited support of the argument to this function (which isn't used in this PR).

prisalcalde:
@dbk78 is this change in the animation looking okay on iOS?

dbk78:
@prisalcalde Good question, I'm not sure how to test the template changes. 😅   I'll get some time in to check it out. 👍

JamieB-gu:
> I'm not sure how to test the template changes

https://github.com/guardian/mobile-apps-article-templates#running-on-ios-simulator

dbk78:
Thanks @JamieB-gu ! Had some issues setting up the environment so I'm getting a hand this morning. 👍

dbk78:
I can see this working the same as Max's gif without the scroll on iOS too. 👍  🙂 

Not sure if this is a known issue, one thing I have noticed on both platforms that happens on both this and the existing "scroll" way of doing things is that the resulting scroll position isn't always correct - I can find my scroll position to be too "high" up the page compared to the target position. It seems the further down the page the target component is, the more "out" the position is. When you scroll back to the top of the live blog and tap the same link again, the position it moves to is 100% correct on the component.

prisalcalde:
That's true @dbk78 . I've just replicated the wrong position on the current implementation. @maxspencer @frankie297 @JamieB-gu any thoughts on the wrong position? Should we create another ticket for this?

maxspencer:
@dbk78 thank you for testing it on iOS, I'll just make the changes suggested by @JamieB-gu and get this merged - thanks Jamie!

@prisalcalde @dbk78 the wrong position is caused by embedded things like tweets that load after the rest of the content and move the blocks around. They're part of the reason we put in the pause that you mentioned Priscilla - to get them to load before scrolling to avoid ending up in the wrong place - but it's not 100% reliable. I expect we could come up with a better solution though!

ArticleTemplates/assets/js/modules/quiz.js
maxspencer: Ah I was copying line 464 and not engaging brain. This way might work but I'll switch to `getElementById` so it's an exact replacement for `#quiz-scores` :+1:

dbk78:
Thanks @maxspencer , that makes sense - it's pretty hard to judge the scroll position when there's dynamically loaded content involved. 🙂

prisalcalde:
Thanks for the explanation, @maxspencer. Makes sense.

PR #1372, opened 2021-04-13
Migrate From `node-sass` To `sass`

JamieB-gu:
## Why?

It looks like the `node-sass` project [is deprecated](https://github.com/sass/node-sass#node-sass) in favour of [Dart SASS](https://sass-lang.com/dart-sass). This PR migrates us over to the JS implementation, which exists in the `sass` package on npm.

The migration was relatively painless. `sass-loader`, which is how we integrate SASS in this project, already supports the `sass` package out of the box. The only difficulty was that we appeared to be using an unsupported pattern in the dark mode stylesheets: `@import` statements inside media queries. `node-sass` seems to allow this through but the new JS implementation doesn't.

I've switched to wrapping each of the dark mode stylesheets in a mixin and using `@include` instead of `@import`. I don't know whether this is the simplest way to do this - if anyone has any other suggestions please let me know!

**Note:** The diff looks excessive, but in reality I've just indented most of these files and wrapped their content in `{}` and a mixin declaration. It might be easier to view the diff in the CLI or VSCode, which does a better job of showing this than the GitHub UI for some reason.

## Changes

- Removed `node-sass` and added `sass`
- Stopped relying on `@import` statements inside media queries
- Switched to mixins for dark mode styles


frankie297 (approved):
Is Apps Rendering the only place that uses `node-sass`?

PR #1395, opened 2021-05-12
Live: 2399 - Dead contributor links not working with `cut_out image`-present

dskamiotis:
Contributors Link not working specifically when a cut_out image is present on the comment article.

The `cutout_image` layer is in front of the `headline` tag and therefore not allowing us to click on the contributor anchor.
The cut out `<div>` is covering the whole top area.

- Finally decided to use `pointer-events:none`
- Adding css property pointer-events: none; to the #other-div will let clicks or other pointer-related events to pass through the div and reach both the canvas and the container

**_For Reference_**

1. Initial investigation:

| Cut_Image Layer with bg-color | Headline layer  and cut_out image with z-index of 100 |
| --- | --- |
|||

2. Initially amended z-index of cut_out image to 50 - however chose not to do this due to potential layer conflicts

| Amended z-index to 50 |
| --- | 
||

- **Final Solution includes `pointer-events: none` (tested in xcode and android studio)**

| Opinion Article | Result of clicking contributor |
| --- | --- |
|||

### DIFF
--- a/ArticleTemplates/assets/scss/modules/_cutout.scss
+++ b/ArticleTemplates/assets/scss/modules/_cutout.scss
@@ -42,6 +42,7 @@
     }
 
     .cutout__image {
+        pointer-events: none;
         background-repeat: no-repeat;
         background-position: bottom right;
         background-size: auto 160px;


JamieB-gu (approved):

marjisound:
Good job @dskamiotis

PR #1399, opened 2021-05-13
Square First Ad Slots

JamieB-gu:
## Why?

We'd like to put square ads in the first ad slot.

## Changes

- Made the first ad slot on the page square

### DIFF
--- a/ArticleTemplates/assets/scss/modules/_advert-slot.scss
+++ b/ArticleTemplates/assets/scss/modules/_advert-slot.scss
@@ -145,7 +145,10 @@
     }
 
     .advert-slot__wrapper--1 {
-        height: 300px;
+        height: 333px;
+        width: 320px;
+        margin-left: auto;
+        margin-right: auto;
     }
 
     .advert-slot__wrapper.test__banner {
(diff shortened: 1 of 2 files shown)


ArticleTemplates/assets/scss/modules/_advert-slot.scss
```diff
@@ -144,6 +144,10 @@
         }
     }
 
+    .advert-slot__wrapper--1 {
+        height: 300px;
+    }
```
ab-gnm, line 149: Thanks for initiating this @JamieB-gu !

The below css worked for me when displaying the advert at the agreed _320px x 320px_ size. The extra height is to account for the thin 'Advertisement' and 'AdChoices' banner within the advert content.

```suggestion
    .advert-slot__wrapper--1 {
        height: 333px;
        width: 320px;
        margin-left: auto;
        margin-right: auto;
    }
```

| Square advert restricted to 320x320 | Non-Teads advert in first slot |
|---|---|
| square | Screenshot_20210521-174442 |

cc: @petternitter @shesah

ab-gnm (approved):
Thanks for the changes, @JamieB-gu . I'd suggest waiting for confirmation from Aoife, here or in chat, before merging.

PR #1421, opened 2021-06-15
Square Ads Only On Narrower Devices

JamieB-gu:
## Why?

We want to serve square ads on narrower devices, and rectangular ads on wider devices. This PR uses a media query to apply the square dimensions only on narrower breakpoints.

Will need confirmation from @ab-gnm @aoifemcl15 @petternitter on whether the chosen breakpoint to switch between the two sizes is correct.

## Changes

- Wrapped square ad CSS in a media query

### DIFF
--- a/ArticleTemplates/assets/scss/modules/_advert-slot.scss
+++ b/ArticleTemplates/assets/scss/modules/_advert-slot.scss
@@ -144,10 +144,14 @@
         }
     }
 
+    // The first ad slot is square on narrower breakpoints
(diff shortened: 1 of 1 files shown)


ab-gnm:
@JamieB-gu Do you know where in this repo can we check the breakpoint at which the ads are split into a separate column for tablets?

JamieB-gu:
I _think_ it's probably `col3`, which is what I've implemented in the code.

ab-gnm (approved):

PR #1438, opened 2021-08-16
[LIVE-2684] Remove width constraint on first ad slot

ab-gnm:
**Jira ticket**: https://theguardian.atlassian.net/browse/LIVE-2671

This change allows us to show non-square ads as full width on Android. The square ads are still limited by the container height specified here (344px) and the ratio constraints in the native code.

Need to confirm this change works for iOS.

| | Before | After |
| --- | --- | --- |
| Horizontal ad | Screenshot_20210816-180834|Screenshot_20210816-174643|
| Square ad |Screenshot_20210816-180711|Screenshot_20210816-181225|


aoifemcl15 (approved):
I've tested on iOS and it doesn't affect our implementation

PR #1526, opened 2022-03-07
Update package.json

akash1810:
| Before | After |
| --- | --- |
|||

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.4.4",
     "classlist-polyfill": "^1.2.0",
-    "curl": "cujojs/curl#0.8.13",
+    "curl": "github:cujojs/curl#0.8.13",
     "d3": "^7.1.0",
     "d3-shape": "^3.1.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -65,7 +65,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.4.4",
     "classlist-polyfill": "^1.2.0",
-    "curl": "cujojs/curl#0.8.13",
+    "curl": "github:cujojs/curl#0.8.13",
```
q00t, line 68: Was just reading https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies

Am wondering if this should be `git+https` ? ie

`git+https://isaacs@github.com/npm/cli.git#0.8.13`

akash1810: I think we should just install from [npm](https://www.npmjs.com/package/curl-amd) like any other node dependency. Curiously the latest version on npm is 0.8.12 whereas 0.8.13 is available on GH releases 🤔 .

akash1810:
This is fixed in #1535. Closing.

PR #1530, opened 2022-04-05
[LIVE-3944] Fix tablet now then interactives sizing bug

michaelclapham:
Fix bug in now and then interactive where on tablet the now photo is taller than the before photo and half the image is cropped (width ways).

The fix for this is very simple. By setting a max width of 570px for the interactives container, this prevents the container from reaching a width of 575px at which point something within the now/then interactive code causes the second image to have a width of 635px (about the width of the iFrame it is contained within). Because this is a max width is has no effect on devices with less wide displays/windows (e.g phones).

## Android Tablet
| Before | After |
| --- | --- |
|||

## Android Phone
| Before | After |
| --- | --- |
|||

## After on Android Tablet and iOS Tablet
Screenshot 2022-04-12 at 16 47 06

### DIFF
--- a/ArticleTemplates/assets/scss/layout/_article--shared.scss
+++ b/ArticleTemplates/assets/scss/layout/_article--shared.scss
@@ -52,7 +52,7 @@
         .element-interactive {
             width: auto;
 
-            // Needed to solve tablet bug in interactive-now-and-then-embed
+            // Prevents uneven sizes of now-and-then images in interactive embeds on Android tablets
             max-width: 570px;
 
             @include inverted-colors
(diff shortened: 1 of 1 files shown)


ArticleTemplates/assets/scss/layout/_article--shared.scss
```diff
@@ -52,6 +52,9 @@
         .element-interactive {
             width: auto;
 
+            // Needed to solve tablet bug in interactive-now-and-then-embed
```
ab-gnm, line 55: We could give more detail here. Something like "This prevents uneven sizes of now-and-then images in interactive embeds on Android tablets".

ab-gnm (approved):
Thanks for the fix, Michael! And good work learning to build with local apps templates.

Before you merge this, please ask someone from iOS to check with this branch to ensure this doesn't degrade the experience for them.
I'd also recommend sharing the tablet before & after images on the email/chat thread so editorial have an understanding of the change we're planning to make here.

michaelclapham:
Tested change on iOS and it stills works as before.

PR #1535, opened 2022-04-20
Migrate `fence` and `curl` dependencies from GitHub to `npm`

ab-gnm:
The PR updates the existing `fence` and `curl` dependencies to be provided via `npm` instead of directly from Github.

#### Why

Something changed over the weekend so that Github now doesn't provide node dependencies without authenticated pulls. The ssh keys for `guardian-android` that we use on TeamCity aren't available inside the Docker container where the Android app is built. As a result, the build is unable to fetch the two dependencies that are provided from github - `fence` and `curl`. This is causing all Android builds to fail on TeamCity.

Moving to npm-hosted versions of these dependencies will help resolve the issue by removing the need for a gihub recognised ssh key inside the docker container.

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/common.js
+++ b/ArticleTemplates/assets/js/bootstraps/common.js
@@ -1,5 +1,5 @@
 import { attach } from 'fastclick';
-import { render } from 'fence';
+import { render } from '@guardian/fence';
 import {
     getClosestParentWithTag,
     debounce,
--- a/package.json
+++ b/package.json
@@ -64,13 +64,13 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.4.4",
+    "@guardian/fence": "^0.2.12",
     "classlist-polyfill": "^1.2.0",
     "curl": "cujojs/curl#0.8.13",
     "d3": "^7.1.0",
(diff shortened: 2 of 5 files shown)


mohammad-haque:
@ab-gnm code changes looks good to me, but I haven't test it locally and probably MSS are best placed to have a final look.

Do we still release updated version of these libraries? Just asking in case we do release updated version then we need to make sure they get to npm as well.

ab-gnm:
> @ab-gnm code changes looks good to me, but I haven't test it locally and probably MSS are best placed to have a final look.
> 
> Do we still release updated version of these libraries? Just asking in case we do release updated version then we need to make sure they get to npm as well.

Thank Mohammad! I tested the changes locally, but I would also prefer someone from MSS to have a look.

Last material update to `Fence` was in [September 2013](https://github.com/guardian/fence/commit/b3d1f4877c18226861af238db0cf74acd1100fd8). 

Last update to `Curl` was in [2016](https://github.com/cujojs/curl/commits/master).

frankie297 (approved):

aoifemcl15 (approved):
I have tested this on iOS and can see template content rendering as expected including live blogs and embeds. 👍

PR #1537, opened 2022-04-25
Reduce height of immersive header background images

DavidLawes:
## Why

On web the height of immersive header background images was reduced (https://github.com/guardian/dotcom-rendering/pull/4671). This change ensures consistency across mobile too.

Note that, because of the top margin calculation, specifying 80vh in these templates resulted in a much reduced background image size. I thought specifying 90vh more closely matched the output in web.

## iOS

| Before | After |
| --- | --- |
|||

## Android

| Before | After |
| --- | --- |
|||

### DIFF
--- a/ArticleTemplates/articleTemplateImmersiveHeight.html
+++ b/ArticleTemplates/articleTemplateImmersiveHeight.html
@@ -1,1 +1,1 @@
-height: calc(100vh - __BODY_OFFSET_TOP__px);
+height: calc(90vh - __BODY_OFFSET_TOP__px);
--- a/ArticleTemplates/assets/js/modules/immersive.js
+++ b/ArticleTemplates/assets/js/modules/immersive.js
@@ -103,7 +103,7 @@ function onImmersiveScroll() {
 }
 
 function getImageHeight() {
-    const viewPortHeight = document.documentElement.clientHeight;
+    const viewPortHeight = document.documentElement.clientHeight * 0.9;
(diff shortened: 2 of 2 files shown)


ArticleTemplates/articleTemplateImmersiveHeight.html
```diff
@@ -1 +1 @@
-height: calc(100vh - __BODY_OFFSET_TOP__px);
\ No newline at end of file
+height: calc(90vh - __BODY_OFFSET_TOP__px);
```
DavidLawes, line 1: iOS substitutes this value when rendering the template (so does android, i think, but android uses more specific styling that ends up overriding this)

ref: https://github.com/guardian/ios-live/blob/master/GLA/GLA/Classes/GLAArticleHtmlRenderer.m#L1476

ArticleTemplates/assets/js/modules/immersive.js
```diff
@@ -103,7 +103,7 @@ function onImmersiveScroll() {
 }
 
 function getImageHeight() {
```
DavidLawes, line 105: Only [android](https://github.com/guardian/mobile-apps-article-templates/blob/master/ArticleTemplates/assets/js/modules/immersive.js#L146) uses this function (for a reason I don't quite understand)

This styling is applied directly to the element (the header background image) - we need this change here to change the styling on android.

frankie297 (approved):

HarryFischer:
Looks good – thanks @DavidLawes !

PR #1539, opened 2022-04-26
add blinking dot in liveblog kicker

marjisound:
This PR adds a blinking dot to liveblogs article title section in order to better indicate to users the liveness on a live blog. 

The accessibility was also tested both in ios and android, and the blinking dot is with no pulsation when the motion/animation is switched off in the device accessibility setting. 

## ios light mode
| Before | After |
| --- | --- |
| image | ezgif-1-f22ab65e00 |

## ios dark mode
| Before | After |
| --- | --- |
| image | ezgif-2-0ed1b7ca31 |

## android dark mode
| Before | After |
| --- | --- |
| image | ezgif-2-6a0c2b35f3 |

## android light mode
| Before | After |
| --- | --- |
| image | ezgif-2-20b8a7edab |

### DIFF
--- a/ArticleTemplates/assets/scss/themes/darkMode/_darkModeLiveblog.scss
+++ b/ArticleTemplates/assets/scss/themes/darkMode/_darkModeLiveblog.scss
@@ -18,6 +18,10 @@
                 background-color: $dark-1;
             }
 
+            .pulsing-dot {
+                color: $dark-3;
+            }
+
             .article-kicker__highlight a {
                 color: $dark-3;
             }
--- a/ArticleTemplates/assets/scss/type/_live.scss
+++ b/ArticleTemplates/assets/scss/type/_live.scss
@@ -36,6 +36,12 @@
             }
         }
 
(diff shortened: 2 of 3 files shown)


JamieB-gu:
Can we make sure this respects the user's accessibility preferences with regards to motion? I'm not sure whether there are other mechanisms in place to do this in the old templates, but if not then this media query might be helpful: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

marjisound:
> Can we make sure this respects the user's accessibility preferences with regards to motion? I'm not sure whether there are other mechanisms in place to do this in the old templates, but if not then this media query might be helpful: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

Thanks @JamieB-gu for your suggestion. I used the media query you showed and tested the blog both in android & ios with motion/animation switched off in accessibility centre and it worked well 🙂

JamieB-gu (approved):

PR #1546, opened 2022-05-05
Fix text size disparity between legacy templates and AR

DavidLawes:
## What does this change
We had reports of a difference in the text size between articles rendered via AR compared to legacy templates. Given we're about to start asking people to pay for the app, this kind of difference isn't really acceptable.

We are updating the legacy templates to ensure the body text (`p` tags) font size matches that of AR articles. Readers can scale the font size on the apps so we need to ensure comparable font sizes across this range. 

On legacy templates the font size in the standfirst and byline do not scale. This feature was added for AR. This feature is not going to be back-ported to the legacy templates and so is excluded from this change.

The body text of AR articles resolve to the following font sizes:

| Smallest | Default | Largest |
| --- | --- | --- |
| 13.6px | 17px | 27.2px |

### Android

| Before | After |
| --- | --- |
|||

| --- | Before | After |
| --- | --- | --- |
| Smallest | 12.8px | 13.6px |
| Default | 16px | 17px |
| Largest | 25.6px | 27.2px |

### iOS

| Before | After |
| --- | --- |
|||

| --- | Before | After |
| --- | --- | --- |
| Smallest |  |  |
| Default | 16px | 17px |
| Largest |  |  |

### DIFF
--- a/ArticleTemplates/assets/scss/base/_global.scss
+++ b/ArticleTemplates/assets/scss/base/_global.scss
@@ -5,6 +5,6 @@ html {
 }
 
 body {
-    font-size: 1.6em;
+    font-size: 1.7em;
     line-height: 1.5em;
 }


JamieB-gu (approved):

DavidLawes:
Hi @ajbreuer - we exchanged comments on a jira ticket about updating the font sizes on articles rendered using the legacy templates. This change reconciles the difference in font size between AR and legacy templates. Let me know if you think this looks ok :)

PR #1596, opened 2022-11-01
LIVE:4733: Fix duplicate videos appearing on iOS

frankie297:
This PR makes sure to concat the list of videos correctly, so when two videos appear with the said ID it renders the video in the correct position. 

The videos were being added to the beginning of the list rather than the end, and causing issues on render. The javascript did not know what position to render the video on the page and so appeared duplicated.

Here, we just make sure the concatenation of the list happens first, ensuring videos are appended in the correct order. So iOS can use the correct order of videos to show to the user. 

| Before | After |
| --- | --- |
|||


aoifemcl15 (approved):
LGTM, thanks for fixing 👍

PR #1604, opened 2022-11-16
Fixed missing kicker in live blogs

tkgnm:
This fixes a bug where the kicker was missing from some live blogs. On line 41 in the` liveBlogTemplate.html` file we are displaying the `__SERIES__` as the kicker. The problem is, not every live blog will be part of a series. In these cases we need to revert to displaying the `__SECTION__` so that we are in line with the website. 

Jamie Byers has mentioned that there are some slightly more complicated rules around what to display as the "kicker" on Dotcom and therefore this change isn't necessarily "perfect". However, it should still cover the large majority of cases and will mean there is no more blank space where there should be a kicker. In any case, it probably isn't worth spending too much time on as the Apps Rendering rollout will make these templates obsolete in the near future. 

| Before | After |
| --- | --- |
|||

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/liveblog.js
+++ b/ArticleTemplates/assets/js/bootstraps/liveblog.js
@@ -111,7 +111,7 @@ function liveblogNewBlockDump() {
 }
 
 function decideKicker() {
-    if (document.getElementsByClassName('article-kicker__highlight') !== undefined) {
+    if (document.getElementsByClassName('article-kicker__highlight')[0] !== undefined) {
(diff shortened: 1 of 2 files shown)


idrise (approved):
With a nice function name to describe intent. Nice.

marjisound (approved):
Great work, just had a question but feel free to merge :)

marjisound:
@tkgnm Did you have a reason to put this js within the html rather than the liveblog.js or another relevant js file?

sophie-macmillan (approved):

marjisound (approved):
Great work Tom 👏

PR #1625, opened 2023-04-04
Callouts

abeddow91:
The EdEx team have built out a new callout component across the AR and DCR rendering platforms. 

Some articles, however, are still rendered using legacy templates. In this circumstance, we see the legacy callout component which is missing the message us functionality. 

This PR updates the CSS and JS for callouts to more closely align the design of callouts in the legacy templates to those from the modern rendering platforms. The html for this can be found [here](https://github.com/guardian/mobile-apps-api/pull/2440).

https://user-images.githubusercontent.com/20416599/229814720-80c0ca62-73de-4692-9d38-cf5da1a864d3.mov

| Before | After |
| --- | --- |
|||


benwuersching (approved):
This looks good, thank you

PR #1626, opened 2023-04-13
Live 3595 remove premium taster test code

tkgnm:
This removes the premium taster code. These are now not needed and their implementation has been removed on both the iOS and Android app. 

JIRA ticket: https://theguardian.atlassian.net/browse/LIVE-3595

| Before | After |
| --- | --- |
|||


DavidLawes (approved):
👍

PR #1653, opened 2023-08-16
Add a block to legacy templates for native apps to insert article player UI

tkgnm:
For en example of how this code would work with the iOS app, see [this branch ](https://github.com/guardian/ios-live/compare/main...test-listen-to-articles)

In the iOS code, calling `articleWebView.evaluateJavaScript("window.setupArticlePlayback()")` will inject a block for the audio player and return to the iOS app a bounding rectangle for where in the viewport to place the player. See screenshot below of a print to the debug console that shows that information.

image

| Before | After |
| --- | --- |
|||


ArticleTemplates/assets/js/bootstraps/common.js
```diff
@@ -59,6 +60,29 @@ function init(liveBlog = false) {
...
 
+function setupArticlePlayback() {
+    // Called by native code
+    window.setupArticlePlayback = addPlaceholderToArticle;
+}
+
+function addPlaceholderToArticle() {
+    const placeholder = document.createElement('div');
+    // we put the audio player just below the standfirst
+    const listenToArticleSibling = document.querySelector('.standfirst');
```
tkgnm, line 71: @waisingyiu so following our meeting with Luke, i guess here instead of querying the `standfirst` we need to find another suitable element to query so that we can place it above the quadrule

michaelclapham:
We could use a way of specifying the height of the blank space from the native side when we call setupArticlePlayback as this can vary depending on the users font size. 

As the user can change this from the toolbar whilst reading the article, we'll also need a way to update the height of the blank space (where we'll write your name).

tkgnm:
Closing this PR as subsequent PR's have now succeeded this work - thanks all

PR #1660, opened 2023-08-23
Migrate build from Team City to Github Actions

tkgnm:
See [this jira ticket](https://theguardian.atlassian.net/browse/LIVE-5541)

This adds a workflow script to build and publish the app templates as an npm package. I tested my changes by going into both the iOS and Android repos, manually changing their templates version numbers, and then running `npm install`. The new package was installed successfully. I then tested the build on the iOS repo and can confirm templates work as expected.

 I paused all team city builds for templates whilst doing this work to ensure that duplicate packages we're not being published simultaneously. 

 You can see the full successful workflow [here](https://github.com/guardian/mobile-apps-article-templates/actions/runs/5955128736)

Screenshot 2023-08-23 at 23 14 27
Screenshot 2023-08-23 at 23 07 48
Screenshot 2023-08-23 at 23 05 25

## Further work

In doing this the iOS/Android repos will no longer automatically pull in the latest apps template updates. A next step would be to try and restore this functionality but I believe that is beyond the scope of this ticket, as Jorge from DevX pointed out (see the jira ticket)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,17 +34,6 @@ jobs:
         if: github.actor != 'dependabot[bot]' && github.ref == 'refs/heads/master' && success()
         runs-on: ubuntu-latest
         steps:
-            - name: Checkout
-              uses: actions/checkout@v2
-
-            - name: Setup Node
-              uses: actions/setup-node@v2
-              with:
-                  node-version: 14
-
-            - name: Install dependencies
-              run: npm install
-
             - name: Set up npm auth token
               run: echo "//registry.npmjs.org/:_authToken=\${NODE_TOKEN}" >> ~/.npmrc
               env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+                  NODE_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+            - name: Bump version and publish if not Dependabot
+              if: success() && github.actor != 'dependabot[bot]'
+              run: |
+                  OFFSET=858
+                  PATCH_VERSION=$((OFFSET + ${{ github.run_number }}))
+
+                  npm --no-git-tag-version version 1.0.${PATCH_VERSION}
+                  npm publish --access public
```
tkgnm, line 61: I think this is neat - we can choose not to publish to npm if the script is being executed by a dependabot pr,  and we can also not bump the workflow if the workflow is unsuccessful which should ensure no gaps in our versions if the workflow is unsuccessful. Let me know if you think this is a good idea!

PR #1677, opened 2023-10-11
Hide and resize height of podcast player

michaelclapham:
Ticket: https://theguardian.atlassian.net/browse/LIVE-5764

Hide existing podcast controls when native player is showing on article.

These provides 2 UX options.
- Place native audio controls in same position as HTML ones were.
- Place native audio controls somewhere else (floating at the bottom of article for example).

We may want to place the native audio player in the same position as the HTML based one, so this function returns it's position relative to the page top left corner. This position does not change as the content is scrolled, keeping it in sync with scrolling is the responsibility of the native app. As the height of a audio player might vary by platform / font size, the height in px can be passed to hidePodcastPlayer as number. e.g hidePodcastPlayer(100).

If the native app just wants to hide the HTML audio controls, it can call hidePodcastPlayer(0)

| Before | After |
| --- | --- |
|||

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/audio.js
+++ b/ArticleTemplates/assets/js/bootstraps/audio.js
@@ -192,6 +192,28 @@ function styleCutoutContainer(duration, cutoutContainer) {
     cutoutContainer.dataset.background = 'true';
 }
 
+/**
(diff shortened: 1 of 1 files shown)


groakland (approved):
LGTM

PR #1679, opened 2023-10-20
LIVE-5793 add button to start mini player

waisingyiu:
According to the design of Podcast articles with a new mini player, a Podcast article has a button (in place of the old audio player control) that starts a mini player to play the podcast.

The button has three states: `Stopped`, `Loading` and `Playing`, which can be set by the native app via Javascript calls.  The functions are attached to `window` object.

The button has an `<a>` element attached with `href` set to a app link.  When the button is in "Stopped" state, the `href` goes to `x-gu://playaudio/<id>`.  When in `Playing` mode, it goes to `x-gu://pauseaudio/<id>`

This PR does not remove the old audio player.  In fact, the template shows the old player by default.  Only when the Javascript function `window.showNewAudioPlayer()` is called, it changes to the new UI.

Paired with @michaelclapham 

| Pillar | Podcast article |
| --- | --- |
| News ||
| Sport ||
| Culture ||
| Lifestyle ||
| Opinion ||


waisingyiu:
@luke-wilkins It would be great if you could check the screenshots to see if they look good to you.

And I verified (via web inspector) that the font weight of the label "Listen to this podcast" has been set to `500` (as stated in the figma file), but the webview of the app (running on my local simulator) does not really make it bold.  (I tried `600` and the app can show it in bold).  Please let me know if it is fine or if you want to change the font weight.

luke-wilkins:
Looks great @waisingyiu - and yes, could we change the 'Listen to podcast' copy to 600 weight please.
We may change the colour of the wave graphic down the line, but all good for now. Thanks.

waisingyiu:
> Looks great @waisingyiu - and yes, could we change the 'Listen to podcast' copy to 600 weight please. We may change the colour of the wave graphic down the line, but all good for now. Thanks.

Thank you @luke-wilkins -  I've changed to font weight to 600 and the screenshots have been updated.

groakland (approved):
LGTM - Approved

PR #1680, opened 2023-10-25
Add follow and notificaiton buttons

lindseydew:
This PR adds in the notification button and changes the svg for the follow button. 
We won't show both states of the notification button at the same time, but this is just to demonstrate what the buttons look like. 

follow_from_article_1

With following selected.
follow_from_article_2


ArticleTemplates/assets/scss/modules/_alerts.scss
```diff
@@ -43,40 +43,57 @@
...
+    &__state--follow  {
         display: none;
     }
 
-    &.following {
-        .alerts__state--follow {
-            display: block;
+    &__state--follow, &__state--notifications-on {
+        svg {
+           fill: white;
```
vlbee, line 79: I think you need to use the article background colour here. not always white. in fact, it might never be actually white.

vlbee: This will be apparent in Comments which are pinkish and maybe darkmode?

vlbee: The colour palette system here is completely different to Source but imagine can match by the hex code. 

These are the colour files:
https://github.com/guardian/mobile-apps-article-templates/blob/5357690df3b8479b7bbeec8d35f102365ba11625/ArticleTemplates/assets/scss/base/_colour.scss

It looks to me like all the colours are being mixed in by class name via specific colour stylesheets:
https://github.com/guardian/mobile-apps-article-templates/tree/5357690df3b8479b7bbeec8d35f102365ba11625/ArticleTemplates/assets/scss/pillar
Darkmode themes:
https://github.com/guardian/mobile-apps-article-templates/tree/5357690df3b8479b7bbeec8d35f102365ba11625/ArticleTemplates/assets/scss/themes/darkMode
Special Tones:
https://github.com/guardian/mobile-apps-article-templates/blob/5357690df3b8479b7bbeec8d35f102365ba11625/ArticleTemplates/assets/scss/tones/_tone--special.scss

vlbee:
The svg icons look a touch high withing the circles to me but this might be the zoom level of the specific screengrab. Some times they shift around I find.

PR #1681, opened 2023-10-26
LIVE-5793 Fix the colour of button in loading state in dark mode

waisingyiu:
I saw that the new play button on podcast article template (created in #1679 ) in *loading* state still showed the old yellow colour under dark mode.  It is expected to have the same colour as the button in stopped / playing states.  

The reason is that in dark mode, the legacy template has another CSS rule to set the colour of the button in loading state, which overrides the rule which is used for the colour in light mode.

This PR fixed the colour by adding a rule for dark mode to override the previous CSS rule.

I took this fix as part of the previous ticket #1679 

| Before | After |
| --- | --- |
| Sport (dark) | Sport (dark) |
|||
| News (dark) | News (dark) |
|||
| Lifestyle (dark) | Lifestyle (dark) |
|||
| Culture (dark) | Culture (dark) |
|||
| Opinion (dark) | Opinion (dark) |
|||

### DIFF
--- a/ArticleTemplates/assets/scss/pillar/_arts.scss
+++ b/ArticleTemplates/assets/scss/pillar/_arts.scss
@@ -13,6 +13,7 @@
         color(arts-inverted),
         color(arts-dark-1),
         color(arts-dark-2),
-        color(arts-dark-3)
+        color(arts-dark-3),
+        color(arts-media-main),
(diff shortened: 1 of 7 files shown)


groakland (approved):
LGTM

PR #1686, opened 2023-11-23
LIVE-6006 Fix the colour of play button when it is being pressed 

waisingyiu:
It was reported that the colour of the play button in the new podcast audio control was incorrectly changed to yellow when it was being pressed.  Further examination showed that it happened under dark mode only.

With the web inspector, it was found that when the button was pressed, the CSS rule with the pseudo class `:active` (an example was shown below) was applied instead of the intended rule:

Screenshot 2023-11-23 at 17 09 06

To fix the problem, new CSS rules have been defined for the colour of play button (in active or not) under dark mode:

Screenshot 2023-11-23 at 17 41 03
Screenshot 2023-11-23 at 17 41 16

The SCSS applies to all pillars.  I captured the screenshots for news and sport pillar as examples

| Before | After |
| --- | --- |
| News | News |
| Pressed | Pressed |
|||
| Sport | Sport |
| Pressed | Pressed |
|||

### DIFF
--- a/ArticleTemplates/assets/scss/themes/darkMode/_darkModeMedia.scss
+++ b/ArticleTemplates/assets/scss/themes/darkMode/_darkModeMedia.scss
@@ -55,6 +55,13 @@
             }
         }
 
+        .audio-player__container_new .audio-player__button.touchpoint--primary .touchpoint__button {
+            background-color: $media-main;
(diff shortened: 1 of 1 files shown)


groakland (approved):
LGTM

PR #1690, opened 2023-12-12
Update Readme to help iOS simulator users

ddramowicz:
No matter how hard I tried, I could not get the app to use the correct version of the templates from a branch.
Finally I tried deleting both `package-lock.json` and `node_modules` - this seemed to do the trick to force npm to install the right versions (and regenerate those files). Maybe this will be helpful for any future users of this.


waisingyiu (approved):
👍

PR #1699, opened 2023-12-15
LIVE-6058 Add listen to article UI to football article template

waisingyiu:
This PR aims to add the new listen-to-article UI to football template.

The football template actually uses the standard article template to render the main part of the article, which we've already added the HTML snippet for the UI.  However, the Javascript functions for controlling the UI were not available in this football template.

This PR extracts these functions into a separate module JS so that it can be shared between article JS and football JS modules. 

| Light | Mode |
| --- | --- |
|||

### DIFF
--- a/.changeset/swift-tigers-listen.md
+++ b/.changeset/swift-tigers-listen.md
@@ -1,0 +1,5 @@
+---
+"@guardian/mobile-apps-article-templates": patch
+---
+
+Add listen-to-article UI to football template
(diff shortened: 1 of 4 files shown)


waisingyiu:
Hi @luke-wilkins , this time it is football template.  It would be great if you can take a look.  Thank you!

luke-wilkins:
All good - thanks

groakland (approved):
LGTM

PR #1703, opened 2024-01-25
Display buttons as inline-block

lindseydew:
This changes the display property of notifications button from `block` to `inline-block` because the issue with setting this to `block` means that the area of the whole region is clickable. 
[This video](https://github.com/guardian/mobile-apps-article-templates/assets/1202622/deb655f0-c5bb-49ed-9db7-c9e7349e3fb4) shows the problem. 

The original intent of using the display property `block` was to put the addition of a second button on a new line. This is now achieved by adding a break tag in the mark up instead. 

followbuttons

### DIFF
--- a/.changeset/stale-dodos-join.md
+++ b/.changeset/stale-dodos-join.md
@@ -1,0 +1,5 @@
+---
+"@guardian/mobile-apps-article-templates": patch
+---
+
+changes the display property of notifications button from block to inline-block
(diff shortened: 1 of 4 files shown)


waisingyiu (approved):
The change looks good to me.  I wonder if it should require a new version of package to be published?  If so, please make a changeset and commit it to the repo.  (instruction is available in readme, just in case)

lindseydew:
> The change looks good to me. I wonder if it should require a new version of package to be published? If so, please make a changeset and commit it to the repo. (instruction is available in readme, just in case)

Thanks yeah I'll make a changeset PR to follow so this can be released

mxdvl (approved):

PR #1710, opened 2024-03-19
Use svg for the circles

lindseydew:
## Why are you doing this?

There was an issue reported that the alignment of the follow icons in android with respect to the circle were off: 
misalignment

The current implementation displays the icon and uses css to add the circle. 

To fix the issue I have changed the image to include svgs that have the icon and the circle within them.


waisingyiu (approved):
It looks great!

Just out of interest, where did you get these SVGs, or did you create them by yourselves?

lindseydew:
> It looks great!
> 
> Just out of interest, where did you get these SVGs, or did you create them by yourselves?

Thanks, they were sent over to me by design

PR #1716, opened 2024-05-20
Support scrollies

mxdvl:
Enable safe scrollies, which do not break out of their `iframe`s.

## Screenshots

https://github.com/guardian/mobile-apps-article-templates/assets/76776/3cbd2628-95f3-4298-b6a6-f9806c333994

## References

- https://github.com/guardian/dotcom-rendering/pull/10604
- https://github.com/guardian/interactive-generic-ai2html-scrolly/pull/6

### DIFF
--- a/ArticleTemplates/assets/js/bootstraps/atoms.js
+++ b/ArticleTemplates/assets/js/bootstraps/atoms.js
@@ -15,6 +15,9 @@ function init() {
     if ('chart' in atomTypes) {
         initCharts();
     }
+    if ('interactive' in atomTypes) {
+        initInteractives();
+    }
 }
 
 function initCharts() {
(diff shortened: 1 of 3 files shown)


ArticleTemplates/assets/scss/modules/_atoms.scss
```diff
@@ -4,3 +4,12 @@
     border: 0;
     margin: 1em 0;
 }
+
+
+iframe.scrolly {
+		position: sticky;
+		top: 0;
+		height: 100vh;
+		height: 100dvh;
```
waisingyiu, line 13: I am not familiar with CSS rule.  May I ask what we set the `height` attribute twice with different values here?  Is it to support old version of browser (webview) which may not support one of them?

ArticleTemplates/assets/js/bootstraps/atoms.js
```diff
@@ -79,4 +82,45 @@ function initExpandables(atom) {
...
+        const iframe = container.querySelector('iframe');
+        let scroll = -rect.top;
+        if (rect.top > 0) {
+            scroll = 0;
+        } else if (rect.top < -rect.height) {
+            scroll = rect.height;
+        }
+        iframe?.contentWindow?.postMessage({
+            kind: 'interactive:scroll',
+            scroll
```
waisingyiu, line 98: Where will this message be handled?  Is it in the interactive atom?
Will it trigger the handler in line 114 below?

waisingyiu (commented):
Thank you @mxdvl for improving the scrolling in interactive pieces in legacy template.

Just a couple of questions because I'm not really familiar with JS stuff.

ArticleTemplates/assets/scss/modules/_atoms.scss
mxdvl: Yes, exactly – CSS is fault-tolerant by default so if the browser [does not support](https://caniuse.com/mdn-css_types_length_viewport_percentage_units_dynamic) the [dynamic viewport units](https://developer.mozilla.org/en-US/docs/Web/CSS/length#dynamic), it will ignore that rule, and fall back to the previous `height` rule that it can parse, which is the non-dynamic version `100vh`.

ArticleTemplates/assets/js/bootstraps/atoms.js
mxdvl: It will be handled inside the `iframe`, [by the interactive atom](https://github.com/guardian/interactive-generic-ai2html-scrolly/pull/6/files#diff-209ac0f7130fd8396a870601ee59b5423e341e90de81626d0ca553013ac0e884R71-R91).

The `interactive:scroll` handler in this file is in response to the atom’s own messages.

waisingyiu (approved):
Thank you @mxdvl for the great work!

